PR #73, opened 2022-01-02
Support for a config_update kwarg when making a Fireworks workflow/firework

arosen93:
## Summary
Currently, it is a bit awkward to update the `manager_config` dictionary (e.g. the `_fworker`) for a bunch of jobs in a flow. In my opinion, it'd be nice to be able to set this in a kwarg when calling `flow_to_workflow()` (and `job_to_firework()` while we're at it). That is what this PR does.

With this PR, doing `flow_to_workflow(flow,  config_update = {"_fworker": "myfworker"})` will update all the jobs in the flow to have `job.config.manager_config["_fworker"] = "myfworker"`.

The same logic applies for `job_to_firework()`.

I'm not sure if you'd like having this functionality, but it is something that I was looking for. Tests have been added.

## Examples
```python
from jobflow.managers.fireworks import flow_to_workflow, job_to_firework

fw = job_to_firework(
    job,
    config_update = {"_fworker": "fworker1"}
)

wf = flow_to_workflow(
    flow,
    config_update = {"_fworker": "fworker1"}
)
```


utf:
Hi @arosen93,

I thought about this and I think I have a slightly more general solution, implemented in #133. Specifically, I added a new `update_config` function to `Job` and `Flow`. Consider a flow containing two jobs.

```python
from jobflow import job, Flow

@job
def add(a, b):
    return a + b
	
add_job1 = add(5, 6)
add_job2 = add(6, 7)

flow = Flow([add_job1, add_job2])
```

The `config` of both jobs could be updated as follows:

```python
new_config = JobConfig(
   manager_config={"_fworker": "myfworker"}, resolve_references=False
)
flow.update_config(new_config)
```

To only update specific attributes, the `attributes` argument can be set. For example, the following will only update the "manager_config" attribute of the jobs' config.

```python
flow.update_config(new_config, attributes="manager_config")
```

Alternatively, the config can be specified as a dictionary with keys that are attributes of the JobConfig object. This allows you to specify updates without having to create a completely new JobConfig object. For example:

```python
flow.update_config({"manager_config": {"_fworker": "myfworker"}})
```

The update_config function also supports a name filter and function filter.

If this still doesn't handle your use case, feel free to re-open this PR.

arosen93:
Beautiful! I look forward to trying this out.

Issue #110, opened 2022-03-15
Docstring Issue: function_filter instead of class_filter


arosen93:
The codeblock below describes `update_maker_kwargs`, which has a `class_filter` but not a `function_filter` kwarg. I think this just needs a find+replace on `function_filter`.

https://github.com/materialsproject/jobflow/blob/fbd5868394c6f4f6b4f2e0ccf4b7ff7d21fe7258/src/jobflow/core/flow.py#L388-L419

utf:
Closed with #111

PR #111, opened 2022-03-15
Fix update_maker_kwargs docs

arosen93:
Docstring fix for #110 and some other instances of the same issue.


utf:
Good catch!

utf:
Good catch!

PR #112, opened 2022-03-15
[WIP] Allow for multiple filters in update_kwargs and update_maker_kwargs

arosen93:
## Summary
Currently, the `name_filter`, `function_filter`, and `class_filter` kwargs in the `update_kwargs` and `update_maker_kwargs` functions of jobs and flows only allow for a single filter. For complex flows in particular, this can become quite cumbersome if you want to update the kwargs of several underlying jobs or makers (but not all of them).

This PR makes it possible to pass a list of filters (in addition to just a single filter) so that the user doesn't have to call `update_kwargs` or `update_maker_kwargs` several times in succession.

### DIFF
--- a/src/jobflow/core/flow.py
+++ b/src/jobflow/core/flow.py
@@ -332,7 +332,7 @@ class Flow(MSONable):
         self,
         update: Dict[str, Any],
         name_filter: Optional[str | List[str]] = None,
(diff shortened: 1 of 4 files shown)


arosen93:
This ended up being more awkward to implement than I would have liked. Not difficult, but awkward. I think I've settled on just keeping it as-is for now.

PR #113, opened 2022-03-15
Update divmaker test

arosen93:
It bothered me that the `DivMaker` in the unit test was an addition operator instead of a division operator. Now updated. Functionally, this does absolutely nothing to the existing unit tests, but... *shrugs*


utf:
Haha, I appreciate the pedantism. Thanks @arosen93.

PR #116, opened 2022-03-17
[WIP]Add list of hosts to Job document

gpetretto:
## Summary

Following previous discussion, here is a draft for the storage of the list of uuids of the Flows containing a job. The idea is to be able to group jobs belonging to the same Flow when querying the Job store. 

The list of hosts is supposed to be ordered in such a way that the first element is the first Flow containing the Job/Flow, with subsequent elements identifying the increasing containers.

The list is stored in the `hosts` attribute. In principle this could entirely replace the `host`, since this should always be the first element of the `hosts`. I can change that as well, if there are no other specific usage of `host`.
I will implement the tests if the current model of handling `hosts` is approved.

A few other additional changes:
* fix the path to the default config file
* validate nested schemas in pydantic models.
* fix few typos

## TODO (if any)

* tests

### DIFF
--- a/src/jobflow/core/flow.py
+++ b/src/jobflow/core/flow.py
@@ -63,6 +63,8 @@ class Flow(MSONable):
     host
         The identifier of the host flow. This is set automatically when an flow
         is included in the jobs array of another flow.
+    hosts
+        The list of UUIDs of the hosts containing the job.
 
     Raises
     ------
@@ -124,6 +126,7 @@ class Flow(MSONable):
         order: JobOrder = JobOrder.AUTO,
         uuid: str = None,
         host: str = None,
+        hosts: Optional[List[str]] = None,
     ):
         from jobflow.core.job import Job
         from jobflow.core.reference import find_and_get_references
@@ -140,6 +143,7 @@ class Flow(MSONable):
         self.order = order
         self.uuid = uuid
         self.host = host
+        self.hosts = hosts or []
 
         job_ids = set()
         for job in self.jobs:
@@ -173,6 +177,8 @@ class Flow(MSONable):
                 raise ValueError(
                     "jobs array does not contain all jobs needed for flow output"
                 )
+
+        self.add_hosts_uuids(self.uuid)
 
     @property
     def job_uuids(self) -> Tuple[str, ...]:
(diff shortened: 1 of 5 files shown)


utf:
Thanks very much for this, it looks fantastic!

As you mentioned, I think you can remove the `host` attribute and just use the first item in the `hosts` array. For backwards compatibility, you could add a `host` property function that returns the first item automatically.

I have a few questions on your other edits:

> fix the path to the default config file

How come you had to edit this? I think the path should be expanded automatically by monty. It works for me on Mac and linux, were you running into issues?

> validate nested schemas in pydantic models.

I'm very excited by this! I had a stab at this before but couldn't get a robust implementation. I was running into issues with branching schemas, for example if a field has the type `Union[model1, model2]` where model1 and model2 have different fields. How well does your implementation handle these cases? It would be great to have something that can keep track of which path you go down. E.g., if you reference an attribute in model1, then subsequent validations are limited to that path. But this might be more complex than its worth, so I'd be very happy to accept a more general solution that provides some level of validation beyond the first key in the model.

gpetretto:
Good! I will proceed with switching to the `hosts` keys and add the tests. 

Concerning the other two points: 

I had this problem on my previous laptop (linux), that I changed a few weeks ago. I did several tests and it was always failing to retrieve the default file without this modification. At the time I thought it was something that had been overlooked, but I just checked and indeed I do not have this issue in my new laptop. Could it be that I had an outdated version of monty in my environment? I can try to retrieve my old environment and check what the problem was. In any case I think it is safe to revert this change.

Indeed my modification of the validation does not handle cases like `Union[model1, model2]`. In the sense that `Union` will not be recognized as a subclass of `BaseModel`, so no further validation will be performed. I think it would be possible to implement some additional level of validation by handling the `Union` case explicitly. However I can imagine some cases that are difficult or impossible to handle in the current approach.   
For example I have seen that in atomate2 there is a field defined as 
```freq_dependent_diel: Union[dict, FrequencyDependentDielectric] = {}```
how to handle the mix of something that is a generic type and a BaseModel?   
In addition you may have a case where `model1` has an attribute `a` and `model2` an attribute `b`. Considering `m: Union[model1, model2]` as a part of the output of a job the following will still be validated as correct
```
Flow([j], output=[j.output.m.a, j.output.m.b])
```
irrespectively of the way the check is implemented, since the validation of the two attributes is performed separately.  Am I right?
I will make some further tests for these particular cases in my implementation.

In the end it was a poor choice to put everything in the same PR. It would be better to split it up and discuss the issues separately. What do you think?

utf:
> Indeed my modification of the validation does not handle cases like Union[model1, model2]. In the sense that Union will not be recognized as a subclass of BaseModel, so no further validation will be performed.

I see. In my opinion, your current implementation is good solution for the time being. I'd be very happy to accept it as is without trying to solve the branching types, as I can imagine that is really difficult. 

> I had this problem on my previous laptop (linux), that I changed a few weeks ago. 

If you had a problem, I think it's fine to keep this in!

> It would be better to split it up and discuss the issues separately. What do you think?

It's up to you. I'm happy with the other changes being in this PR as long as you include the relevant tests :)

gpetretto:
In the end, while implementing the tests I thought of some more points that may be worth discussing. I will open different PRs to keep the topics clearly separated.

PR #117, opened 2022-03-18
List of hosts in Jobs and Flows

gpetretto:
## Summary

Allowing the storage of the list of uuids of the Flows containing a job. The idea is to be able to group jobs belonging to the same Flow when querying the Job store.

The list of hosts is supposed to be ordered in such a way that the first element is the first Flow containing the Job/Flow, with subsequent elements identifying the increasing containers.

The list is stored in the `hosts` attribute. `host` is removed and replaced by a property giving the first element of `hosts`, if present. 

An additional issue that emerged was that, when using Fireworks as a manager, for restart, detours and additions the new Jobs are always wrapped inside a Flow, leading to clashes with the list of `hosts`. The easiest way that I have found to fix this was to always wrap restart/detour/addition Jobs into a Flow during the `Job.run()`. Is this acceptable?

An additional open point is the fact that the list of jobs inside `Flow` is accessible by the client code without any restriction. This makes impossible to guarantee that the jobs in that list has the correct list of hosts properly set.  This seems also a general issue, since all the checks on the Jobs performed in `__init__` are skipped if someone tampers with the jobs list directly.  
I propose to make the list of `jobs` in Flow not directly accessible and provide an API to modify it through a method of `Flow`, that will also take charge of performing all the required actions to properly add a new job to the list. `jobs` could be either a tuple, or could be stored internally as a list under `_jobs` and the `job` property returns a copy of the list, so that it cannot be directly modified.  
Would you agree to switch to this approach?

### DIFF
--- a/src/jobflow/core/flow.py
+++ b/src/jobflow/core/flow.py
@@ -60,11 +60,9 @@ class Flow(MSONable):
         the order automatically based on the connections between jobs.
     uuid
         The identifier of the flow. This is genenrated automatically.
-    host
-        The identifier of the host flow. This is set automatically when an flow
-        is included in the jobs array of another flow.
     hosts
-        The list of UUIDs of the hosts containing the job.
+        The list of UUIDs of the hosts containing the job. This is updated
+        automatically when a flow is included in the jobs array of another flow.
 
     Raises
     ------
@@ -125,7 +123,6 @@ class Flow(MSONable):
         name: str = "Flow",
         order: JobOrder = JobOrder.AUTO,
         uuid: str = None,
-        host: str = None,
         hosts: Optional[List[str]] = None,
     ):
         from jobflow.core.job import Job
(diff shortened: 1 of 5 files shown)


utf:
Hi @gpetretto, thanks very much for this PR. You changes so far look perfect.

> I propose to make the list of jobs in Flow not directly accessible and provide an API to modify it through a method of Flow, that will also take charge of performing all the required actions to properly add a new job to the list. jobs could be either a tuple, or could be stored internally as a list under _jobs and the job property returns a copy of the list, so that it cannot be directly modified.

I think this makes sense and will prevent users running into strange behaviour. I think my preference is your first suggestion. I.e.,
- jobs is a tuple
- a new `add_job` function to add a new job or list of jobs to a Flow.
- similarly, a new `remove_job` function to remove a job or list of jobs from a flow (specified by the job index).

Both `add_job` and `remove_job` should perform the same checks as the init method.

If you're rather, those changes can be done in a separate PR to keep this one short?

gpetretto:
Hi @utf, thanks for reviewing this. If you approve these changes I think this PR can be merged and I will make the updates to the jobs list in a new one.

utf:
Great, thanks!

PR #118, opened 2022-03-18
Validate subschema

gpetretto:
## Summary

Implement a first level of validation of outputs with nested Schemas.

The current implementation ignores the checks if nested schemas are inside `Union`, `List`, `Dict`, while it correctly validates if inside `Optional`. 

As previously discussed (#116), implementing the checks for `Union` will probably be challenging and it will not be supported for the time being. Should we do the same for `List` and `Dict`?


utf:
This is great, thanks @gpetretto.

> Should we do the same for List and Dict?

Yeah, I think ignoring these cases is best for now.

PR #124, opened 2022-03-22
add s3store yaml parsing

jmmshn:
## Fixed parsing of nested stores (Ex. S3Store)
Now checks the `dict_spec` for the word "type" and recursively calls the constructor.

### DIFF
--- a/src/jobflow/core/store.py
+++ b/src/jobflow/core/store.py
@@ -661,13 +661,14 @@ class JobStore(Store):
 
 
 def _construct_store(spec_dict, valid_stores):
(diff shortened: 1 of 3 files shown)


utf:
Please could you also add an example of this format in the JobflowSettings: https://github.com/materialsproject/jobflow/blob/45eb719f5aec3dab1aaf47b2034069186c6d1069/src/jobflow/settings.py#L44-L58

jmmshn:
Ok ready to merge

utf:
Error is unrelated to this PR: https://github.com/jupyter/nbconvert/issues/1736

utf:
Thanks!

PR #129, opened 2022-03-31
fix failing test

gpetretto:
## Summary

Looks like my PR #117 made some changes that were incompatible with #121, resulting in a failing test. This should fix it.


utf:
Thanks for spotting this.

PR #130, opened 2022-03-31
Typo fixes

mjwen:
## Summary

* Fix `restart.py` example, restart->replace
* A few typo fixes


src/jobflow/core/reference.py
```diff
@@ -61,7 +61,7 @@ class OutputReference(MSONable):
     ----------
     uuid
         The job uuid to which the output belongs.
-    attributes :
+    attributes
```
utf, line 64: I think this was there because otherwise the linting gets confused.

utf:
Thanks 👍

PR #131, opened 2022-04-01
Control changes to job list in Flow

gpetretto:
## Summary

Add functionalities to modify the list of Jobs/Flows in a Flow and prevent direct changes to it.  
`jobs` in `Flow` is a private tuple and should only be modified by the `add_jobs`, `remove_jobs` methods. Also `output` undergoes checks if it gets modified outside the `__init__` to avoid missing references.

I also changed one test of the settings since it was always picking my ~/.jobflows.yaml file and always failing locally.

### DIFF
--- a/src/jobflow/core/job.py
+++ b/src/jobflow/core/job.py
@@ -261,7 +261,8 @@ class Job(MSONable):
     config
         The config setting for the job.
     hosts
-        The list of UUIDs of the hosts containing the job.
+        The list of UUIDs of the hosts containing the job. The object identified by one
+        UUID of the list should be contained in objects identified by its subsequent elements.
     **kwargs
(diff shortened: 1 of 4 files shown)


src/jobflow/core/job.py
```diff
@@ -1052,8 +1053,7 @@ def prepare_replace(
         store_output_job.index = current_job.index + 1
         store_output_job.metadata = current_job.metadata
         store_output_job.output_schema = current_job.output_schema
-        store_output_job.add_hosts_uuids(replace.uuid)
```
utf, line 1055: Should this line have been removed?

utf (commented):
This looks perfect, thank you. Just one comment.

PR #132, opened 2022-04-02
Move `mypy` + `isort` + `flake8` config from `.pre-commit-config.yaml` to `setup.cfg`

janosh:
Same treatment as in https://github.com/materialsproject/atomate2/commit/6270579a615aa333ad0b8348ce8fcd55dadce59a. Plus some typo fixes.

### DIFF
--- a/src/jobflow/managers/local.py
+++ b/src/jobflow/managers/local.py
@@ -165,4 +165,4 @@ def run_locally(
     if ensure_success and not finished_successfully:
         raise RuntimeError("Flow did not finish running successfully")
 
-    return dict(responses)
+    return responses
--- a/src/jobflow/managers/local.py
+++ b/src/jobflow/managers/local.py
@@ -165,4 +165,4 @@ def run_locally(
     if ensure_success and not finished_successfully:
         raise RuntimeError("Flow did not finish running successfully")
 
-    return responses
+    return dict(responses)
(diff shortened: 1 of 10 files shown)


src/jobflow/managers/local.py
```diff
@@ -165,4 +165,4 @@ def _run(root_flow):
     if ensure_success and not finished_successfully:
         raise RuntimeError("Flow did not finish running successfully")
 
-    return dict(responses)
+    return responses
```
utf, line 168: This is actually purposeful to convert from defaultdict to dict. Purely a style thing but I think this should stay as is?

utf (commented):
Thanks! Just one minor comment.

src/jobflow/managers/local.py
janosh: `defaultdict` subclasses `dict` so for the purpose of `mypy`, there's no need to convert. At run time `defaultdict` offers all the methods of `dict` too so no potential for problems there either (afaik). But happy to revert if you prefer.

utf: Yeah, its not for the API, functionality, or mypy. Just purely stylistic when working with jobflow in a jupyter notebook for example. Thanks for reverting.

utf:
Thanks!

Issue #149, opened 2022-05-11
BUG: `jsanitize` should not use `strict=True` when storing job output


arosen93:
**Describe the bug**
When Jobflow tries to store the outputs of a `job`, it runs `jsanitize(response.output, strict=True, enum_values=True`). In my opinion, `strict` should not be set to `True` because this prevents the conversion of "unknown" object types to `str` and instead leads to a crash. For instance, consider the following imaginary numpy array:
```
test =  np.array([1 + 2j, 3 + 4j, 5 + 6j])
```
If this is in the output, `jsanitize` will lead to an error like that shown below rather than simply converting the unknown data types to a string.

A better solution is to instead use `jsanitize(response.output, enum_values=True, recursive_msonable=True)`, which will still call `.as_dict()` for MSONables but not on unknown data types. I added this kwarg a little while ago to `monty`. See additional context [here](https://github.com/materialsvirtuallab/monty/issues/343).

```python
  File "/usr/share/miniconda/envs/test/lib/python3.8/site-packages/jobflow/core/job.py", line 571, in run
    output = jsanitize(response.output, strict=True, enum_values=True)
  File "/usr/share/miniconda/envs/test/lib/python3.8/site-packages/monty/json.py", line 505, in jsanitize
    return {
  File "/usr/share/miniconda/envs/test/lib/python3.8/site-packages/monty/json.py", line 506, in <dictcomp>
    k.__str__(): jsanitize(
  File "/usr/share/miniconda/envs/test/lib/python3.8/site-packages/monty/json.py", line 499, in jsanitize
    return [jsanitize(i, strict=strict, allow_bson=allow_bson, enum_values=enum_values) for i in obj.tolist()]
  File "/usr/share/miniconda/envs/test/lib/python3.8/site-packages/monty/json.py", line 499, in <listcomp>
    return [jsanitize(i, strict=strict, allow_bson=allow_bson, enum_values=enum_values) for i in obj.tolist()]
  File "/usr/share/miniconda/envs/test/lib/python3.8/site-packages/monty/json.py", line 545, in jsanitize
    obj.as_dict(),
AttributeError: 'complex' object has no attribute 'as_dict'
```

utf:
I think of `strict=True` as a feature not a bug. If the output is not valid json then it won't be stored in the database properly and can't be used by other jobs properly.

[...]

PR #150, opened 2022-05-11
[WIP] Don't use `strict=True` on `jsanitize` of job output

arosen93:
Addresses #149.


arosen93:
@utf, why do you want the job to fail if the output isn't MSONable? As a user, I feel like I'd much rather it be converted to a `str` representation than fail outright.

Issue #140, opened 2022-04-22
Enhancement: non informative information get stored into additional stores 


mjwen:
If an additional store is set and then we run atomate2 jobs, like the [tutorial example](https://materialsproject.github.io/atomate2/user/running-workflows.html), non informative info like the screenshot below will be saved there.  

This is because the TaskDocument of atomate2 contains `force_constants` and `normalmode_eigenvecs`, and the two are supposed to be stored in the additional store as specified [here](https://github.com/materialsproject/atomate2/blob/92e6d6b5f33e87ce434f992995b9392eb4348def/src/atomate2/vasp/jobs/base.py#L28). But simple jobs like static and relax ones will not have these info computed from VASP. Then, in jobflow, their values in [object_map](https://github.com/materialsproject/jobflow/blob/c3063bb6d550ace28c4a1fc9bbc88b469ad5ec9f/src/jobflow/core/store.py#L305) will be `None`. Together with the fact that `force_constants` and `normalmode_eigenvecs` given [here](https://github.com/materialsproject/atomate2/blob/92e6d6b5f33e87ce434f992995b9392eb4348def/src/atomate2/vasp/jobs/base.py#L28) are strings but not a class,  [_get_blob_info](https://github.com/materialsproject/jobflow/blob/c3063bb6d550ace28c4a1fc9bbc88b469ad5ec9f/src/jobflow/core/store.py#L767) can not extract any useful info. This gives the non informative info stored in the additional store. 

Probably, we can do some additional check to 1) add more info (e.g. what these keys are) or 2) not store them at all if values are None? Personally I prefer the latter. 

<img width="679" alt="Screen Shot 2022-04-21 at 11 15 25 PM" src="https://user-images.githubusercontent.com/10203513/164614664-348aeabf-ed9a-432c-95d3-e3428b1c8ef7.png">

utf:
Ah thanks for noticing this. I definitely prefer option 2. Would you be happy to add a PR with a test?

mjwen:
Sounds good. Will do.

PR #160, opened 2022-06-01
Fix data store storing None

mjwen:
## Summary

This fixes the JobStore issue that stores `None` in additional store as discussed in #140.  

## Checklist

  type check your code.


utf:
Great, thank you!

PR #165, opened 2022-06-08
searchable fields

jmmshn:
## Make sure that the "job_uuid" and "job_index" can be searched on

Currently, the "job_uuid" is not stored in the index store of the S3Store.
Which means they are not searchable even though they are in the query results.

This should fix that behavior.


utf (commented):
Thanks very much and also nice that you added the test! Just a quick comment.

jmmshn:
oops!  thanks for catching that.
Fixed now.

utf:
Great, thank you!

PR #177, opened 2022-07-10
update_kwargs patch

jmmshn:
# Reworked update_kwargs,

This makes checking and validating maker parameters for complex makers much easier by recycling (with some modification) the code used for update_kwargs.

- Created a general function that can update or validate makers.
- The `recursive_call` function does not actually need to be recursive since the location found by `find_key` already contains all the objects that need to be worked on.
- We just have to reverse sort them by length to make sure we work on the deepest objects first.
- Then we can call a general function on all the Makers in the filter and update them if desired.
- The fact that `location` is sorted will make sure the updates are properly propagated up the chain.

The original PR message was for a smaller fix.
> ## Prevent extras calls to update_kwargs
> 
> I think if `update_kwargs` is called with `nested=True` it will already go in and look for all the makers starting from the top level.
> So when you call `update_kwargs` recursively it does not need `nested=True` anymore.

### DIFF
--- a/src/jobflow/core/job.py
+++ b/src/jobflow/core/job.py
@@ -785,7 +785,6 @@ class Job(MSONable):
                 nested=nested,
                 dict_mod=dict_mod,
             )
-            print(">> ", maker)
             setattr(self, "function", getattr(maker, self.function.__name__))
         elif nested:
             # also look for makers in job args and kwargs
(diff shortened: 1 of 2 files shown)


jmmshn:
@utf this is ready for review and is needed by the atomate2 defects PR.

utf:
Thanks @jmmshn!

PR #183, opened 2022-07-28
Added default for additional_stores in the settings.

davidwaroquiers:
One of the main advantages of jobflow is to be able to start running a workflow (Flow) almost without any configuration, at least in terms of databases to set up (well you still have to configure e.g. vasp, pseudos, ... for atomate2). In atomate2, additional stores are used in the JobStore (e.g. "data"). The default JOB_STORE used a MemoryStore for the "main" store (the JobStore.docs_store) but did not set any default for the additional stores. These additional stores of course are not known in advance (their keys in the dict). For atomate2, it uses "data" but other implementations may use other key(s) for these additional stores. This PR solves this by using a defaultdict with a factory function (lambda) that returns a MemoryStore the first time it is accessed.

## Summary

- Modified the default JobStore so that it handles additional_stores with on-the-fly generated MemoryStore's

### DIFF
--- a/src/jobflow/core/store.py
+++ b/src/jobflow/core/store.py
@@ -328,7 +328,13 @@ class JobStore(Store):
         self.docs_store.update(dict_docs, key=key)
 
         for store_name, blobs in blob_data.items():
-            store = self.additional_stores.get(store_name, None)
+            # Here we use a try/except with a self.additional_stores[store_name]
+            # instead of self.additional_store.get(store_name, None) to allow
+            # for additional_stores to be a defaultdict with a MemoryStore's default.
+            try:
+                store = self.additional_stores[store_name]
+            except KeyError:
+                store = None
 
             if store is None:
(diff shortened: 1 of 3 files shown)


JaGeo:
This would actually help me a lot as I otherwise have to start a MongoDB database on the front end of our supercomputer for every tiny test ...

davidwaroquiers:
> This would actually help me a lot as I otherwise have to start a MongoDB database on the front end of our supercomputer for every tiny test ...

Hi Janine,

You can actually define your jobflow.yaml with the additional data-MemoryStore in it. This PR allows to make this automatically/dynamically and irrespective of the names of the additional stores.

JaGeo:
Thanks. Yes, I know ... It would require, however, much less work for new users as well.

utf:
Thanks for this. I agree that it'll be useful and should be the default behaviour!

Issue #193, opened 2022-08-22
Enhancement: Set metadata in Maker


gpetretto:
Currently it is possible to use the `update_metadata` method to set the metadata in Jobs and Flows. However, a Job can use a Maker to generate new Jobs/Flows at runtime . If I am not wrong there is no way for a user to easily set the metadata for the dynamically generated Jobs. 

I though it would be interesting to start a discussion about introducing a standardized way to deal with such a use case.  
An option could be to add a `metadata` attribute to the Maker base class and suggest that subclasses implementing the `make` method use it to call the update_metadata if they return a new Flow, while the `make` methods that are decorated with `@job` can automatically set the metadata. 

Do you think this may be an interesting feature? And does anybody have a better idea about how to implement it?

utf:
There is some support for setting the metadata of dynamically created jobs. Specially, if you set the metadata of job that replaces itself with a new flow/job, the new flow/job will inherit the metadata of the original job. Whether this approach is flexible enough I'm not sure. Can you think of a situation where the current approach doesn't work but where your suggestion would work?

> suggest that subclasses implementing the make method use it to call the update_metadata if they return a new Flow

I've considered doing something like this before too. I'm a bit worried as it relies on people to actually implement this, and I'm sure many people will forget. Another approach could be to add a decorator that does this automatically, which potentially has a slightly lower barrier to people implementing it.

However, I think this raises a broader worry I have about controlling the settings of dynamic workflows. For example, you might want to have finer control over the job config of dynamically created jobs. The current approach is the same as for metadata (the config just gets passed down to dynamic jobs) but this might not be flexible enough.

gpetretto:
Thanks for looking into this.

Consider the following example 

```python
@job
def number(n):
    return n


@job
def use_maker_job(maker):
    flow = maker.make()
    return Response(replace=flow)


@dataclass
class TestMaker(Maker):
    def make(self):
        j1 = number(1)
        j2 = number(2)
        return Flow([j1, j2], output=[j1.output, j2.output])

job = use_maker_job(TestMaker())
job.update_metadata({"test_meta": "test"})

run_locally(job)
```

This creates 4 entries in the DB, 1 for the initial job, two for the `number` jobs and one for the "replacement" job. Out of these only the first and last have the metadata set.  The same will happen for an addition/detour.
[...]

PR #198, opened 2022-08-31
Delayed updates to config and metadata

gpetretto:
## Summary

Implementing delayed updates for config and metadata as discussed in #193 

Added job name in the stored job document.

Fixing few issues in the update_config/metadata:

- The `function_filter` was working only in the case of instantiated `Job`s and not for `@job` decorated functions. Switched to checking the wrapped functions. 
- The logic for `name_filter` was wrong in case of the job name is `None`. In principle the name is never `None`, but fixed it anyway

### DIFF
--- a/src/jobflow/core/flow.py
+++ b/src/jobflow/core/flow.py
@@ -512,6 +512,7 @@ class Flow(MSONable):
         name_filter: str | None = None,
         function_filter: Callable | None = None,
         dict_mod: bool = False,
+        dynamic: bool = True,
     ):
         """
         Update the metadata of all Jobs in the Flow.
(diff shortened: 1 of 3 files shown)


utf:
This looks great! Thanks very much.

My only comment is on the documentation. This is a complicated feature whose use won't be particularly obvious at first glance.  Do you think you could include a minimal example in the docstring of for `Job.update_config` and `Job.update_metadata` and mention dynamic workflows and the response object?

gpetretto:
I added an explanation in the docstrings. Let me know if you think it needs some improvement.

utf:
Those examples are perfect. Thank you very much.

PR #221, opened 2022-10-05
Allow `Response.output` to be type hinted

mkhorton:
## Summary

A small change to allow `Response.output` to be type hinted.

This should allow method signatures like:

`my_job() -> Response[int]`

or 

`my_job() -> Response[Structure]`

to work and be type checked correctly by `mypy` or similar, while `my_job() -> Response` should still work as previously. The type in the square brackets is the type for the `Response.output` specifically.

Not sure if there's an easier way to do this, but could be helpful?

### DIFF
--- a/src/jobflow/core/job.py
+++ b/src/jobflow/core/job.py
@@ -1098,8 +1098,15 @@ class Job(MSONable):
             self.hosts.extend(hosts_uuids)
 
 
+# For type checking, the Response output type can be specified
+# in a type hint via this type variable.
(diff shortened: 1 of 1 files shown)


mkhorton:
For purposes of review, I found https://mypy-play.net to be very helpful for quick testing.

utf:
This is awesome Matt!

I wonder if there is something else we can add so that type hinting works properly with decorated functions? E.g., some functions may just return a Structure directly but the decorator will mean a Response object is returned. Either way, this is great.

PR #293, opened 2023-03-30
[Bug Fix] Fixed nested data storage bug

jmmshn:
## Summary

This should fix the following bug:
https://github.com/materialsproject/atomate2/issues/291

Currently, there is a problem when you serialize nested data when you have `Child` nested in `Parent`, both will get uploaded to the `additional_store`.  But I think you should ever only serialize the `Parent` in all but a few exceptional cases.

### DIFF
--- a/src/jobflow/core/store.py
+++ b/src/jobflow/core/store.py
@@ -8,6 +8,7 @@ from maggma.core import Store
 from monty.json import MSONable
 
 from jobflow.core.reference import OnMissing
+from jobflow.utils.find import get_root_locations
 
 if typing.TYPE_CHECKING:
     from enum import Enum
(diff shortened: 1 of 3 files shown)


utf:
Thanks for this! I actually tried to prevent nested objects being stored when I wrote the original function. e.g., see here: https://github.com/materialsproject/jobflow/blob/c2a316100ac192fc12cfa03269dda4f871689a12/src/jobflow/utils/find.py#L40-L41 But that assumes the nesting of the same object. It makes more sense that the objects of the same type wouldn't be nested so this was quite an oversight.

I actually think the `_get_root_locations` could be useful elsewhere in the future. Would you mind:
- moving it to `jobflow.utils.find`
- making it a public method (e.g., remove `_` from the beginning of the name)
- adding a docstring/type hinting for the `locations` parameter.

Thanks again @jmmshn, this is very helpful.

PR #310, opened 2023-05-02
[Bug fix] Test imports need to be in function 

jmmshn:
## Test imports need to be in function 

Moved import inside


utf:
Thanks!

PR #331, opened 2023-06-02
Fixed JobStore.from_dict_spec so that the original dict_spec is not modified

davidwaroquiers:
Fixed JobStore.from_dict_spec so that the original dict_spec is not modified.
The spec is now copied so that if this is used programmatically, the type is not removed from the original spec_dict, which may be used afterwards, (e.g. stored in a database), or if there is some validation by a pydantic schema.

@utf For information, this is making the jobflow-remote fail in some of the config cases.

## Summary

- Modification of the internal `_construct_store` method so that the original dict_spec is not modified.

None

## TODO (if any)

None


utf:
Thanks!

PR #320, opened 2023-05-17
[WIP] Update `fw_spec` in Fireworks manager

arosen93:
The passed `fw_spec` was never being added to the job metadata. That is now taken care of.

I am merely a proxy for @samblau who was too lazy to open the PR himself. 😉

PR #345, opened 2023-06-12
Propagate tags from fw_spec to metadata

sivonxay:
Picking up #320.

I think the changed originally proposed would add alot of redundant information into the document/db.

### DIFF
--- a/src/jobflow/managers/fireworks.py
+++ b/src/jobflow/managers/fireworks.py
@@ -155,7 +155,7 @@ class JobFiretask(FiretaskBase):
 
         # Add the metadata from the fw_spec
         fw_tags = fw_spec.get("tags", {})
-        job.metadata.update({"tags": fw_spec["tags"]})
+        job.metadata.update({"tags": fw_tags]})
 
         if hasattr(self, "fw_id"):
             job.metadata.update({"fw_id": self.fw_id})
--- a/src/jobflow/managers/fireworks.py
+++ b/src/jobflow/managers/fireworks.py
@@ -53,10 +53,7 @@ def flow_to_workflow(
     flow = get_flow(flow)
 
     for job, parents in flow.iterflow():
-        fw = job_to_firework(job,
-                             store,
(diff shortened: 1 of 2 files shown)


sivonxay:
After speaking to @samblau, it seems that the functionality desired is to propagate the tags from the fw_spec into the metadata stored in the document.

While the documentation recommends to use `Job.update_metadata({"tags": ["test"]})` function, people who are starting to use jobflow who have previously used fireworks+atomate are used to using the `atomate.common.powerups.add_tags()` function to add the tags. Because the tags aren't added to the metadata, they aren't propagated into the documents by jobflow.

src/jobflow/managers/fireworks.py
```diff
@@ -153,6 +153,11 @@ def run_task(self, fw_spec):
             store = SETTINGS.JOB_STORE
         store.connect()
 
+        # Add the metadata from the fw_spec
+        fw_tags = fw_spec.get("tags", None)
+        if fw_tags is not None:
+            job.metadata.update({"tags": fw_tags})
```
utf, line 159: Maybe we should also check if tags already exist and if they are a list, and in that case append the tags?

utf (commented):
Thanks for this. I just had one minor comment.

src/jobflow/managers/fireworks.py
sivonxay: Alright. Added.

One small issue I encountered here was that tags were getting duplicated if the job metadata had the "tags" key in it. Because of this line:  [ spec.update(job.metadata)  # add metadata to spec](https://github.com/materialsproject/jobflow/blob/1386b113e18fb0faf42992d1e2ed3a379b206123/src/jobflow/managers/fireworks.py#L116). 

To account for this, I just removed all duplicates from `job.metadata["tags"]` assuming that duplicates will never be desired.

utf: That's a good catch. Thanks!

Issue #349, opened 2023-06-14
BUG: to_mermaid


JaGeo:
**Describe the bug**
I tried to run the following code.

```python
from atomate2.vasp.flows.elastic import ElasticRelaxMaker
from jobflow.utils.graph import to_mermaid
from pymatgen.core.structure import Structure

flow=ElasticRelaxMaker().make(Structure.from_file("SbSe/POSCAR"))

print(to_mermaid(flow))
```

It fails with:

```
AttributeError                            Traceback (most recent call last)
Cell In[3], line 7
      3 from pymatgen.core.structure import Structure
      5 flow=ElasticRelaxMaker().make(Structure.from_file("SbSe/POSCAR"))
----> 7 print(to_mermaid(flow))

File /hpc-user/jgeorge/PycharmProjects/2023_06_09_Debug_Phonon_kpath/jobflow/src/jobflow/utils/graph.py:249, in to_mermaid(flow, show_flow_boxes)
    246             if indent_level != 1:
    247                 lines.append(f"{prefix}{job.uuid}({job.name})")
--> 249 add_subgraph(flow)
    251 return "\n".join(lines)

File /hpc-user/jgeorge/PycharmProjects/2023_06_09_Debug_Phonon_kpath/jobflow/src/jobflow/utils/graph.py:236, in to_mermaid.<locals>.add_subgraph(nested_flow, indent_level)
```

JaGeo:
I know that it is probably not intended to just visualize one job but I might not be the only one that by accident passes a single job instead of a Flow.

JaGeo:
I am closing it as the documentation clearly says it should be a Flow.

davidwaroquiers:
A quick fix would just be to check if it's a single Job, and if so, just wrap it with a Flow and continue just like that.

JaGeo:
> A quick fix would just be to check if it's a single Job, and if so, just wrap it with a Flow and continue just like that.

This would work if a single job in a Flow would create an output. Maybe, there is a real bug then:
```
from jobflow.utils.graph import to_mermaid
from jobflow import job, Flow
@job
def add(a, b):
    return a + b
add_first = add(1, 2)
my_flow = Flow(jobs=[add_first])
graph_source = to_mermaid(my_flow)
print(graph_source)
```
produces

```
flowchart TD
```

JaGeo:
[...]

PR #350, opened 2023-06-14
fix mermaid with one job (in a flow)

JaGeo:
## Summary

Closes the issue described in #349 . As described there, I am not sure if I missed a use-case for the following line https://github.com/materialsproject/jobflow/blob/09ffa16026b5682f7947045fedf6376ce7276d7f/src/jobflow/utils/graph.py#L246

I did not see any change in other tested workflow visualizations.


utf:
Thanks! Can't remember why I added that line.

PR #369, opened 2023-07-17
`Flow` + `Job` magic methods

### LINKED ISSUE #368: Flow + Job magic methods
How come `Jobs` and `Flows` don't have 

- `__len__`
- `__getitem__`
- `__setitem__`
- `__iter__`
- `__repr__`

and friends?

Would be nice if `len(flow)` returned an `int` and `print(flow)` gives insightful output about containing each job's name etc.

```py
flow = MPMetaGGARelax().make(structure)

len(flow)
> TypeError: object of type 'Flow' has no len()
# expected 3

print(flow)
> <jobflow.core.flow.Flow object at 0x2c1386190>
# expected:
# MPMetaGGARelax flow consisting of
# 1. MP pre-relax
# 2. MP meta-GGA Relax
# 3. MP meta-GGA Static
```

janosh:
Closes #368.

```py
from pymatgen.core import Lattice, Structure
from atomate2.vasp.flows.mp import MPMetaGGARelax

struct = Structure(
    lattice=Lattice.cubic(3),
    species=("Fe", "Fe"),
    coords=((0, 0, 0), (0.5, 0.5, 0.5)),
)
flow = MPMetaGGARelax().make(structure)

len(flow)
>>> 3
print(flow)
>>> Flow(name='MP Meta-GGA Relax', uuid='d43cd776-fb64-40f0-bacf-1933ee2342b8')
consisting of 3 jobs:
  1. Job(name='MP pre-relax', uuid='99a016af-eb9d-4d7a-8454-85ecf6b59d41')
  2. Job(name='MP meta-GGA Relax', uuid='496006d6-5375-49ce-91a8-cf53483173e1')
  3. Job(name='MP meta-GGA Static', uuid='8fe17841-a536-4d39-9bfa-dd83bc7900a2')
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,8 +2,8 @@ default_language_version:
   python: python3
 exclude: '^src/atomate2/vasp/schemas/calc_types/'
 repos:
-- repo: https://github.com/charliermarsh/ruff-pre-commit
-  rev: v0.0.250
+- repo: https://github.com/astral-sh/ruff-pre-commit
+  rev: v0.0.280
   hooks:
   - id: ruff
     args: [--fix]
--- a/docs/tutorials/8-fireworks.md
+++ b/docs/tutorials/8-fireworks.md
@@ -94,6 +94,16 @@ flow.update_config({"manager_config": {"_fworker": "fworker1"}}, name_filter="jo
 flow.update_config({"manager_config": {"_fworker": "fworker2"}}, name_filter="job2")
 ```
 
+NB: There are two ways to iterate over a `Flow`. The `iterflow` method iterates through a flow such that root nodes of the graph are always returned first. This has the benefit that the `job.output` references can always be resolved.
+`Flow` also has an `__iter__` method, meaning you can write
+
+```py
+for job_or_subflow in flow:
+    ...
+```
+
+to simply iterate through the `Flow.jobs` array. Note that `jobs` can also contain other flows.
+
 ### Launching the Jobs
 
 As described above, convert the flow to a workflow via {obj}`flow_to_workflow` and add it to your launch pad.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -111,6 +111,7 @@ show_missing = true
 exclude_lines = [
     '^\s*@overload( |$)',
     '^\s*assert False(,|$)',
+    'if TYPE_CHECKING:',
     'if typing.TYPE_CHECKING:',
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,54 +11,54 @@ license = { text = "modified BSD" }
 authors = [{ name = "Alex Ganose", email = "alexganose@gmail.com" }]
 dynamic = ["version"]
 classifiers = [
-    "Programming Language :: Python :: 3",
-    "Programming Language :: Python :: 3.8",
-    "Programming Language :: Python :: 3.9",
-    "Programming Language :: Python :: 3.10",
     "Development Status :: 2 - Pre-Alpha",
+    "Intended Audience :: Information Technology",
     "Intended Audience :: Science/Research",
     "Intended Audience :: System Administrators",
-    "Intended Audience :: Information Technology",
     "Operating System :: OS Independent",
-    "Topic :: Other/Nonlisted Topic",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
     "Topic :: Database :: Front-Ends",
+    "Topic :: Other/Nonlisted Topic",
     "Topic :: Scientific/Engineering",
 ]
 requires-python = ">=3.8"
 dependencies = [
-    "monty>=2021.5.9",
-    "pydash",
-    "networkx",
-    "maggma>=0.38.1",
-    "pydantic",
     "PyYAML",
+    "maggma>=0.38.1",
+    "monty>=2021.5.9",
+    "networkx",
+    "pydantic",
+    "pydash",
 ]
 
 [project.optional-dependencies]
 docs = [
-    "sphinx==7.1.1",
-    "furo==2023.7.26",
-    "myst_parser==2.0.0",
-    "ipython==8.14.0",
-    "nbsphinx==0.9.2",
     "autodoc_pydantic==1.9.0",
+    "furo==2023.5.20",
+    "ipython==8.14.0",
+    "myst_parser==2.0.0",
+    "nbsphinx==0.9.2",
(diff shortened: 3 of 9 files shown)


utf:
This is great! Thanks very much.

I notice from the tests that I assumed that job comparison would depend not only on the UUIDs but also on the specific function arguments. I think this is just relying on the default dataclass `__eq__()` operator. See: https://docs.python.org/3/library/dataclasses.html#module-contents

Personally, I think having stricter equality is better approach but I am open to other thoughts.

utf:
I think those are very good points @davidwaroquiers. To add to that, some comments:

1. Flow already has a method `iterflow` which iterates through the flow in such an order that the job output references can always be resolved. E.g., root nodes of the graph are always returned first.
2. The `__iter__` method introduced here has different behaviour and only iterates through the jobs array (which can also contain other flows). This could be somewhat confusing. 
3. If we change the behaviour of `__iter__` and `__len__` to be consistent with subflows, then this would make the behaviour of `__setitem__` and `__getitem__` inconsistent.

Maybe its ok to leave the implementation as is, but I agree that the behaviour should be clarified. It would be nice to add examples of how to use these methods in the docs somewhere.

janosh:
Great comments, thanks @utf @davidwaroquiers.

> This is indeed nice. Just to comment or push the discussion a bit further, what "should" be the len of a Flow if it is made of other Flows ? Could be the number of "steps" (jobs or flows) or the number of "flattened" jobs.

I considered returning the flattened length of a `Flow` but I think the default length should be just the number of items in the `jobs` list.
I think the user should have to ask for the flattened length explicitly. We could add a property

```py
@property
def flat(self):
    ...
```

which would recursively unpack subflows. Then people can use `len(flow.flat)`.

> Same for the print, what happens when you'd print a Flow with Flows in it.

@davidwaroquiers That's already handled actually. If `__str__` is undefined, Python automatically falls back on `__repr__` (not vice versa) and since `Flow.__repr__` puts every `subflow` in `f"{subflow}"` which calls `__str__` on it which calls `__repr__` on that subflow, it's already recursive. I just tweaked `__repr__` to increase indentation for every level down in the subflow hierarchy for added readability. Now this prints:

```py
from jobflow import Flow, Job


def add(a, b):
    return a + b

def get_test_job():
    return Job(add, function_args=(1, 2))

subflow1 = Flow([get_test_job()])
subflow2 = Flow([get_test_job(), get_test_job()])
subsubflow = Flow([get_test_job(), get_test_job()])
subflow3 = Flow([get_test_job(), get_test_job(), subsubflow])
print(Flow([subflow1, subflow2, subflow3]))
```

```py
Flow(name='Flow', uuid='f7ae544a-7e8d-4976-b768-13e431199fa6') consisting of 3 jobs:
  1. Flow(name='Flow', uuid='9e7b6459-adca-48b1-8c00-b86e15876a38') consisting of 1 jobs:
    1.1. Job(name='add', uuid='0d4a4621-3249-4f94-bdb1-a83910982921')
  2. Flow(name='Flow', uuid='d2b1cdbe-0075-4bba-8686-3851fe81a4b0') consisting of 2 jobs:
    2.1. Job(name='add', uuid='bf687f6b-3e5b-4e48-b579-e1646b0707c7')
    2.2. Job(name='add', uuid='9422bbf9-8c91-4153-a19b-5aad36d96431')
  3. Flow(name='Flow', uuid='b62ebaaf-bea2-489f-8e88-75a2eebfa7a9') consisting of 3 jobs:
    3.1. Job(name='add', uuid='5fae9b50-8051-438e-b295-14298569c527')
    3.2. Job(name='add', uuid='1dbba533-4cc7-4cef-be76-d9d773cfaf80')
    3.3. Flow(name='Flow', uuid='b6526c18-ce3c-48c2-ae26-6207859ccb46') consisting of 2 jobs:
      3.3.1. Job(name='add', uuid='c0c2eb67-65ab-4bcc-a4cc-465dd8898f2c')
      3.3.2. Job(name='add', uuid='42e8f5d2-42d1-4885-a627-13232362447d')
```

I'll have to get back to the other comments later.

janosh:
Looking at it now, I guess we can drop the `consisting of n jobs` part.

janosh:
Just had time to take another look at this.

#### `__eq__`

I'm a bit torn about `Job.__eq__` on the one hand I agree two flows should be different if they have different attributes. On the other hand, what's the point of a UUID if it doesn't uniquely identify a flow? UUID stands for [universally unique identifier](https://en.wikipedia.org/wiki/Universally_unique_identifier).

@utf Btw, `Jobs` are not dataclasses so before they had the default Python behavior of only being equal if they're literally the same object in memory.

https://github.com/materialsproject/jobflow/blob/9928fdbb3ee6191668e89ff33273f6abc3defcd7/src/jobflow/core/job.py#L219-L221

#### `iterflow`

I actually think it's a plus to have two ways of iterating through a `Flow`. As you said @utf, we just need to be clear in the docs on how `for job in flow.iterflow` and `for job in flow` differ. But I can see use cases for either one. `pandas` also has many different ways of iterating though a dataframe (`df.iterrows`, `df.items`, `df.itertuples`, `for col in df`, `df.keys` and maybe more).

#### `__repr__`

@davidwaroquiers The problem with `__repr__` kwargs like `depth: int` is people usually aren't aware of them. They just print the object or return it from a Jupyter cell in which case Python/`ipykernel` implicitly calls `__repr__` on that object without passing any arguments. So I don't think there would be much value in a `depth` keyword. To use it would mean

```py
print(flow.__repr__(depth=2)
```

Happy to take further suggestions esp. about `__eq__` but for now I'll start adding tests and docs for the implementation as is.

utf:
On the `__eq__` point, due to how some of the dynamic behaviour works (specifically the replace functionality), multiple jobs can have the same `uuid` but different `index`s. 

So maybe for `__eq__` the check could just be whether both uuid and index match.

janosh:
@utf Other than the `__eq__` decision, I think this is in a good state and ready to go.

utf:
Thanks @janosh. I'm ok with the current `__eq__` implementation but we could always switch it in a future PR if there is a strong need.

I don't mean to be a stickler, but is it possible to add a few more tests so that the test coverage doesn't downgrade. Part of the issue is that you have changed:
```python
if typing.TYPE_CHECKING
```
to
```python
if TYPE_CHECKING
```
in some cases, and that means the codecov is moaning about not checking those lines. The simple fix is to add the latter to the coverage exception list here:
https://github.com/materialsproject/jobflow/blob/5602cfd0b4d93ac58184e86039a0c213750ec608/pyproject.toml#L111-L115

But there are a couple of `if` statements etc that are not currently covered.

janosh:
On it!

janosh:
@utf Coverage is back up to 100%. 🎉

utf:
Brilliant, thank you!

PR #375, opened 2023-07-21
Allow bson objects as job inputs, e.g. datetime.datetimes

mcgalcode:
## Summary

Adds the `allow_bson=True` kwarg to the `jsanitize` invocation in `contains_job_or_flow` utility function, and a corresponding test case. I recently tried to pass an object containing a `datetime.datetime` in it as an input to a job (specifically, the object was an emmet `TaskDoc`. This object was the output of another job (and I had retrieved it from the jobstore).

I noticed that every other invocation of jsanitize in the codebase uses `allow_bson=True`, with the exception of the one in `Job::as_dict`, so I concluded that this might be a bug. Please feel free to close if this behavior is intentional! However, I think it's useful to allow things like `datetime`s as inputs (at least to allow the use of TaskDocs as input parameters).

## Checklist

title.

- [X] Docstrings have been added in the[Numpy docstring format](
- [X] Type annotations are **highly** encouraged. Run [mypy](http://mypy-lang.org/) to
  type check your code.
- [X] Tests have been added for any new functionality or bug fixes.

### DIFF
--- a/src/jobflow/utils/find.py
+++ b/src/jobflow/utils/find.py
@@ -213,7 +213,7 @@ def contains_flow_or_job(obj: Any) -> bool:
         # argument is a primitive, we won't find an flow or job here
         return False
 
-    obj = jsanitize(obj, strict=True)
+    obj = jsanitize(obj, strict=True, allow_bson=True)
 
     # recursively find any reference classes
     locations = find_key_value(obj, "@class", "Flow")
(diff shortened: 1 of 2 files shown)


janosh (approved):
Thanks @mcgalcode for the fix! And extra credits for the test! 🙏 😄

PR #409, opened 2023-08-29
Fix issue with data store

### LINKED ISSUE #393: How to store a part of the output of a Flow in a Response in a data store
I am currently struggling to find a way to store a part of the output of Flow, that is  created with a Response object, in the data store. Is this possible in general? 
If I can pass it via "stored_data": could we then document it in a better way?

JaGeo:
## Summary
Closes #393 

@utf , I have now tested your suggestion from #393 . It works. Everything gets stored in the datastore.

Any idea for a test? I will try to come up with something but I am still not very good with jobflow itself as decorators are not exactly my strength.

### DIFF
--- a/src/jobflow/core/job.py
+++ b/src/jobflow/core/job.py
@@ -1332,7 +1332,6 @@ def prepare_replace(
         metadata = replace.metadata
         metadata.update(current_job.metadata)
         replace.metadata = metadata
-        replace._kwargs = current_job
 
         if not replace.output_schema:
             replace.output_schema = current_job.output_schema
--- a/src/jobflow/core/job.py
+++ b/src/jobflow/core/job.py
@@ -1332,6 +1332,7 @@ def prepare_replace(
         metadata = replace.metadata
         metadata.update(current_job.metadata)
         replace.metadata = metadata
+        replace._kwargs = current_job._kwargs
 
         if not replace.output_schema:
             replace.output_schema = current_job.output_schema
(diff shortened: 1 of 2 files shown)


JaGeo:
@utf, should hopefully now be ready to merge.

src/jobflow/core/job.py
```diff
@@ -1331,6 +1332,7 @@ def prepare_replace(
         metadata = replace.metadata
         metadata.update(current_job.metadata)
         replace.metadata = metadata
+        replace._kwargs = current_job._kwargs
```
utf, line 1335: I'm not sure about propagating the `_kwargs` for a single job replace.

For example, if you have the following setup, I would expect the data store of `job2` to be used, rather than job1.

```python
@job
def job1():
    replaced_job = job2()
    return Response(replace=replaced_job)

@job(data=["output"])
def job2():
    return {"output": my_big_data}
```

To give you a concrete example, if you have a job which updates the input set of a vasp calculation and then submits a replace, you would expect the data settings for the vasp job to be used rather than the initial job.

```python
def update_input_set(structure, vasp_maker):
    vasp_maker.input_set_generator.use_incar_settings["NSW"] = 100
    vasp_job = vasp_maker.make(structure)
    return Response(replace=vasp_job)
```

In this implementation, the data store settings of the `vasp_job` will be removed.

I think copying over the kwargs is only needed for the last `store_inputs` job of a replaced flow, since otherwise it is impossible to set it. For a single replaced job, you can just set the data store settings yourself.

utf (commented):
Thanks @JaGeo, just one comment.

src/jobflow/core/job.py
JaGeo: I see! Thanks. I will remove this and then it should hopefully be ready to merge!

utf:
Thanks very much for the fix.

Issue #411, opened 2023-08-30
BUG: A blank `~/.jobflow.yaml` file causes a crash


arosen93:
If the user has a `~/.jobflow.yaml` file and it's blank, the following error is raised:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/rosen/software/miniconda/envs/quacc/lib/python3.10/site-packages/jobflow/__init__.py", line 14, in <module>
    SETTINGS = JobflowSettings()
  File "pydantic/env_settings.py", line 39, in pydantic.env_settings.BaseSettings.__init__
  File "pydantic/main.py", line 342, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 1 validation error for JobflowSettings
__root__
  'NoneType' object is not iterable (type=type_error)
```

mcgalcode:
What do you think of leaving a warning for the user? https://github.com/materialsproject/jobflow/pull/415

arosen93:
@mcgalcode: I think it would be cleaner if the presence of a blank YAML simply was interpreted as `{}` and didn't update any of the default parameters.

mcgalcode:
I could raise a RuntimeError or SyntaxError for a malformed file and do nothing or leave a warning for empty an empty file.

arosen93:
@mcgalcode: I think something like that might be a good solution! I don't personally know the best error to raise (that always confuses me...).

mcgalcode:
Alright, updated the PR to warn if it's empty and raise a `ValueError` with a more descriptive message if it's badly formatted.

arosen93:
Cool! Thanks! Once merged, we'll probably need to make the same change to Atomate2 😄

utf:
Fixed in #415

PR #415, opened 2023-08-31
Add warning when jobflow.yaml is blank or badly formatted

mcgalcode:
## Summary

Addresses #411, and adds moto as a testing dependency.

- moto is added as a test dependency, but this is because it is already used in a test in tests/core/test_store.py

## Checklist

title.

- [X] Tests have been added for any new functionality or bug fixes.
- [X] All linting and tests pass.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,7 +45,7 @@ docs = [
     "sphinx==7.2.4",
 ]
 dev = ["pre-commit>=2.12.1"]
-tests = ["pytest-cov==4.1.0", "pytest==7.4.0"]
+tests = ["pytest-cov==4.1.0", "pytest==7.4.0","moto==4.2.0"]
 vis = ["matplotlib", "pydot"]
 fireworks = ["FireWorks"]
 strict = [
--- a/src/jobflow/settings.py
+++ b/src/jobflow/settings.py
@@ -1,5 +1,6 @@
 """Settings for jobflow."""
 
+import warnings
 from collections import defaultdict
 from pathlib import Path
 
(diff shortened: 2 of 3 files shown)


src/jobflow/settings.py
```diff
@@ -136,10 +137,21 @@ def load_default_settings(cls, values):
...
-            new_values.update(loadfn(config_file_path))
+            if Path(config_file_path).stat().st_size == 0:
+                warnings.warn(
+                    f"An empty JobFlow config file was located at {config_file_path}"
+                )
+            else:
+                try:
+                    new_values.update(loadfn(config_file_path))
+                except ValidationError:
+                    raise ValueError(
```
utf, line 150: Strangely codecov is is saying this that this line isn't being tested. Do you have any ideas?

Could this be because pydantic is actually raising a `ValueError` to begin with, so this except is not being hit?

pyproject.toml
```diff
@@ -45,7 +45,7 @@ docs = [
     "sphinx==7.2.4",
 ]
 dev = ["pre-commit>=2.12.1"]
-tests = ["pytest-cov==4.1.0", "pytest==7.4.0"]
+tests = ["pytest-cov==4.1.0", "pytest==7.4.0","moto==4.2.0"]
```
utf, line 48:
```suggestion
tests = ["pytest-cov==4.1.0", "pytest==7.4.0", "moto==4.2.0"]
```

utf (commented):
Thanks very much. Only thing is why the test isn't actually catching the value error.

src/jobflow/settings.py
mcgalcode: Haha - I saw this last night and couldn't figure it out, but I looked again and I think you're right. It's confusing because the printed stacktrace says `ValidationError`:

```
pydantic/env_settings.py:40: in pydantic.env_settings.BaseSettings.__init__
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

>   ???
E   pydantic.error_wrappers.ValidationError: 1 validation error for JobflowSettings
E   __root__
E     dictionary update sequence element #0 has length 1; 2 is required (type=value_error)

pydantic/main.py:341: ValidationError
```

But the error you have to catch is `ValueError`.

Anyway, I switched to `ValueError` and made an assertion about the message.

utf:
Thanks very much!

PR #424, opened 2023-09-08
Formalizing the JobStore document format as a pydantic model

hrushikesh-s:
## Summary

Include a summary of major changes in bullet points:

- Adding the functionality to parse the JobStore document as a `pydantic model` in place of `dictionary`

## Checklist

Work-in-progress pull requests are encouraged, but please put [WIP] in the pull request
title.

Before a pull request can be merged, the following items must be checked:

- [X] Code is in the [standard Python style](https://www.python.org/dev/peps/pep-0008/).
  The easiest way to handle this is to run the following in the **correct sequence** on
  your local machine. Start with running [black](
  https://black.readthedocs.io/en/stable/index.html) on your new code. This will
  automatically reformat your code to PEP8 conventions and removes most issues. Then run
  [pycodestyle](https://pycodestyle.readthedocs.io/en/latest/), followed by [flake8](
  http://flake8.pycqa.org/en/latest/).
- [X] Docstrings have been added in the[Numpy docstring format](
  https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html).
  Run [pydocstyle](http://www.pydocstyle.org/en/2.1.1/index.html) on your code.
- [X] Type annotations are **highly** encouraged. Run [mypy](http://mypy-lang.org/) to
  type check your code.
- [X] Tests have been added for any new functionality or bug fixes.
- [X] All linting and tests pass.

Note that the CI system will run all the above checks. But it will be much more
efficient if you already fix most errors prior to submitting the PR. It is highly
recommended that you use the pre-commit hook provided in the repository. Simply
`cp pre-commit .git/hooks` and a check will be run prior to allowing commits.

Issue #374, opened 2023-07-20
Insufficient information stored to recover parent-child relationship between jobs from their `JobStore` output docs?


mkhorton:
Please advise if I've mis-interpreted the code/docs.

Assume:

* Storing `Job` outputs via `JobStore`.
* Not using a workflow manager, i.e. using `jobflow` directly.

For a given document in the `JobStore`, I see `uuid`, I also see `hosts` (which can be used to see that a given `Job` belongs to the same `Flow`), however, that I can see, there is no way to see the dependency relationship between two or more `Job` output documents, is this correct?

If correct, is this intended usage? What would be a minimal way to retain this information, without adding a dependency on a specific workflow manager?

utf:
Hi @mkhorton, this is something I've spoken to @gpetretto and @davidwaroquiers about. I believe the only other information you need to resolve the job dependencies is the `OutputReferences` in the job inputs. These are available through the `job.input_references` property.

The simplest way to enable this would be:
1. At the beginning of the `job.run` function, copy the output of `job.input_references`. The reason why we have to copy them at the beginning is that the `job.resolve_args` function resolves the references in place. So at the end of the function the original input references are not available.
2. Add a new field field `"input_references" to the data stored at the end of `job.run`. E.g., here: https://github.com/materialsproject/jobflow/blob/53f0c76e4cc0baf40c1f9365b31ad2604cc8a601/src/jobflow/core/job.py#L599

You should then be able to construct the entire flow (including nested flows) and the dependencies between jobs. The only information that will be missing is the names of the Flows (the names of the jobs are fine). The reason is that we don't store flows in the database directly.

mkhorton:
Thanks for the reply @utf, good to know I wasn't missing anything obvious.

I'll see if I can make a PR to add this, unless @gpetretto or @davidwaroquiers are already working on it? If it'd be welcome, I'd quite like to add a `pydantic.BaseModel` to describe the JobStore document format too.

utf:
A PR would be very welcome. And yes, agreed that we should have a document model for the job store document.

davidwaroquiers:
[...]

PR #425, opened 2023-09-09
Reconstruct flow from outputs in JobStore [WIP]

mcgalcode:
## Summary

This is a WIP PR addressing #374 . I implemented the storage of input references in the JobStoreDocument (from Hrushikesh's PR ([here](https://github.com/materialsproject/jobflow/pull/424)).

## Checklist

title.

- [X] Docstrings have been added in the[Numpy docstring format](
- [X] Type annotations are **highly** encouraged. Run [mypy](http://mypy-lang.org/) to
  type check your code.

### DIFF
--- a/src/jobflow/schemas/job_store.py
+++ b/src/jobflow/schemas/job_store.py
@@ -4,6 +4,8 @@ import typing
 
 from pydantic import BaseModel, Field
 
+from jobflow.core.reference import OutputReference
+
 
 class JobStoreDocument(BaseModel):
     """A Pydantic model for Jobstore document."""
@@ -31,4 +33,7 @@ class JobStoreDocument(BaseModel):
     name: str = Field(
         None,
         description="The name of the job.",
+    )
+    input_references: typing.List[OutputReference] = Field(
+        default_factory=list, description="The list of input references for this job"
     )
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,7 +40,7 @@ docs = [
     "furo==2023.8.19",
     "ipython==8.14.0",
     "myst_parser==2.0.0",
-    "nbsphinx==0.9.2",
+    "nbsphinx==0.9.3",
     "sphinx-copybutton==0.5.2",
     "sphinx==7.2.4",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,7 +51,7 @@ fireworks = ["FireWorks"]
 strict = [
     "FireWorks==2.0.3",
     "PyYAML==6.0.1",
-    "maggma==0.53.0",
+    "maggma==0.53.1",
     "matplotlib==3.7.2",
     "monty==2023.8.8",
     "moto==4.1.14",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,7 +54,7 @@ strict = [
     "maggma==0.53.1",
     "matplotlib==3.7.2",
     "monty==2023.8.8",
-    "moto==4.1.14",
+    "moto==4.2.0",
     "networkx==3.1",
     "pydantic==1.10.9",
     "pydash==7.0.6",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,7 +42,7 @@ docs = [
     "myst_parser==2.0.0",
     "nbsphinx==0.9.2",
     "sphinx-copybutton==0.5.2",
-    "sphinx==7.2.2",
+    "sphinx==7.2.4",
 ]
 dev = ["pre-commit>=2.12.1"]
 tests = ["pytest-cov==4.1.0", "pytest==7.4.0"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,7 +45,7 @@ docs = [
     "sphinx==7.2.4",
 ]
 dev = ["pre-commit>=2.12.1"]
-tests = ["pytest-cov==4.1.0", "pytest==7.4.0"]
+tests = ["pytest-cov==4.1.0", "pytest==7.4.0","moto==4.2.0"]
 vis = ["matplotlib", "pydot"]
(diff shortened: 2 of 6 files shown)


mcgalcode:
I think there are a lot of possible interaction patterns people will use for retrieving and navigating flow outputs, so I just sketched some out here. I'd like to get some feed back from @utf and @arosen93 to see if any of this looks good.

I'm happy to rework/remove any of this per consensus here, but just wanted to throw something out there for a first pass.

mcgalcode:
The interface of `FlowOutput` ([here](https://github.com/materialsproject/jobflow/pull/425/files#diff-aaa35c931a6edf303b26f27bce1f9d51464d670ec851fe88f7164607731141d0R172)) sort of sketches out the interaction patterns I was imagining.

davidwaroquiers:
Hi @mcgalcode ,


Very nice that you start this WIP PR and great start! I did not get into all the details, but from what I understand of the implementation, the Flow can be reconstructed if its jobs have finished (as job output documents are inserted into the database only when the complete). Do I understand this correctly? I guess it would be nice to reconstruct a Flow even when it hasn't started or while some of its jobs have completed and some others are running (or waiting), or be able to reconstruct a flow with a job that has failed (the failed job won't appear in the reconstructed Flow I think, as there won't be any job document to it).

Any ideas ?

mcgalcode:
Hi @davidwaroquiers ! Sorry for the delayed response here, I think I missed turning on notifications for this one.

You understand correctly. This code reconstructs a flow from output objects that are present in the main output document store.

I like the idea of being able to reconstruct flows that have yet to be started, or flows that are incomplete or partially failed, but I'm a little hazy on how that would work. For instance, if a flow hasn't started yet (i.e., it hasn't been run by some type of `manager`), I thought it doesn't exist anywhere aside from the memory of the program that instantiated it. Is that not the case?

I can imagine doing this in the case of a _particular_ manager, i.e. the fireworks manager. In that case, I think utilities for reconstructing a Flow from it's representation in fireworks would be particularly useful, and I actually have some sketchy helper functions that I use for this in my own work. Is that what you mean?

I think my jobflow understanding may be a bit insufficient here :)

davidwaroquiers:
No you are perfectly right. I am actually always thinking in terms of an ongoing development you are not yet aware of (normal it's in private repos). I will add you to the repos (it's a remote execution mode of jobflow, which ultimately might be included directly into jobflow itself). If you have questions about it, feel free to contact me and I can give you a few more details.

src/jobflow/schemas/job_store.py
```diff
@@ -0,0 +1,39 @@
...
+    """A Pydantic model for Jobstore document."""
+
+    uuid: str = Field(
+        None, description="A unique identifier for the job. Generated automatically."
+    )
+    index: int = Field(
+        None,
+        description="The index of the job (number of times the job has been replaced.",
+    )
+    output: typing.Any = Field(
```
mkhorton, line 20: Instead of `typing.Any`, this could be made a Generic:

```
T = TypeVar("T")

class JobStoreDoc(BaseModel, Generic[T]):

    ...

    output: T
```

This would allow people to easily sub-class when they're working with a JobStore document with a specific output type.

mkhorton: On this note, I had a similar document model implemented that I had intended to upstream. One addition I found helpful was:

```
    @validator("output", pre=True)
    def reserialize_output(cls, v):
        if isinstance(v, dict) and "@module" in v and "@class" in v:
            v = MontyDecoder().process_decoded(v)
        return v
```

Note this is in pydantic v1 syntax, I think it will have to be modified slightly for pydantic v2.

This is useful in the scenario that you're loading a dict directly, e.g. `JobFlowDocument(**dict)`, rather than using MontyDecoder which I think would take care of it automatically.

mcgalcode: Thanks Matt! These are good suggestions. We have another PR that this one depends on that isolates the implementation of that model right here: https://github.com/materialsproject/jobflow/pull/424. This one is a clunky use of git.

@hrushikesh-s Do you want to update your PR with these suggestions?

mcgalcode:
@davidwaroquiers Aha! Thanks for this clarification, I'll take a look at the repos you added me to. Sounds like it will inspire your suggestion :)

I do think that the functionality you're talking about would be very useful. It can be a little confusing to interact with job outputs since there is no formal record of whatever flow they belong to.

src/jobflow/schemas/job_store.py
hrushikesh-s: @mkhorton thanks for the suggestions. 
@mcgalcode , yes I'll update my PR accordingly.

hrushikesh-s: @mkhorton: Jobflow uses Pydantic v1 (see [here](https://github.com/materialsproject/jobflow/blob/main/pyproject.toml#L34)). Is there any plan to upgrade it to v2 anytime soon? If not then your implementation of `@validator("output", pre=True)` should work as it is, and we don't need to change the syntax. 
Maybe I'm missing out on something there?

Andrew-S-Rosen: I plan to migrate Jobflow to pydantic2 very very soon. I am waiting on Jason to patch emmet and maggma first, which he said should be done in a day or two.

hrushikesh-s: @Andrew-S-Rosen thanks for the update. For the time being, I've implemented the `validator` as per `pydantic v1` syntax in #424, but changing it to `pydantic v2` should be straightforward. I'll modify it once your migration from `pydantic v1` to `v2` is complete.

Andrew-S-Rosen: Sounds good! Thanks. Will be good to keep tabs on what needs switching and where.

PR #467, opened 2023-10-29
Simplify code by using `Flow` methods `__len__` and `__getitem__`

janosh:
Saves accessing `flow.jobs` a bunch.


utf:
Hi @janosh, please could you revert this change, since it overlaps with the `__getitem__` stuff proposed in #450. The `__len__` part should still be fine.

janosh:
could you specify what you mean by "have to rewrite the Flow.__getitem__"? do you want to make it behave like a dict rather than a list of jobs?

PR #477, opened 2023-11-07
Allow SSHTunnel in job stores 

mjwen:
## Summary

Allow the use of [SSHTunnel](https://github.com/materialsproject/maggma/blob/86a161ec6a8eb05b8b78a48fc083ba0688f22dd0/src/maggma/stores/mongolike.py#L42) in job stores. This will make it possible to use a `jobflow.yaml` like the below 
```yaml
  docs_store:
    type: MongoStore
    ssh_tunnel:
      type: SSHTunnel
      ...
  additional_stores:
    data:
      type: S3Store
        ...
      index:
        ...
      s3_profile:
        ...
      ssh_tunnel:
        type: SSHTunnel
        ...
```

[MongoStore](https://github.com/materialsproject/maggma/blob/86a161ec6a8eb05b8b78a48fc083ba0688f22dd0/src/maggma/stores/mongolike.py#L128) already have ssh tunnel support. Support for S3Store is coming from this PR https://github.com/materialsproject/maggma/pull/882

The `SSHTunnel` class is simply added to the `all_stores` variable. Technically, `all_stores` no longer only contains stores. So, we might want to change the variable name.


mjwen:
Test fails because SSHTunnel is relocated in this PR https://github.com/materialsproject/maggma/pull/882 

Once that is merged, we should be good.

utf:
Thanks!

PR #481, opened 2023-11-13
JOSS: Add more complete Zenodo reference info

Andrew-S-Rosen:
Addresses https://github.com/openjournals/joss-reviews/issues/5995#issuecomment-1808396589.

Namely, I previously used a minimal format for the Zenodo references. Now, I have used the one auto-generated by Zenodo. One downside of this approach is that the author list includes all contributors to the repo, no matter how large or small their contributions are.

I see two ways to go about this:

1) Just accept it and move on.

2) Keep the rest of the metadata but remove the `author` field.

@utf, any thoughts?


utf:
Happy to just accept it!

PR #494, opened 2023-11-23
Add fibonacci example to tutorial

JaGeo:
@utf , I simply added the example to the tutorial as well. This is more a result of me being unable to look beyond the documentation (and being to tired to think clearly), but maybe others also overlook the example folder.


utf:
Thanks!

PR #499, opened 2023-11-28
fix Flow deserialization for hosts

gpetretto:
## Summary

There was an issue with the (de)serialization of a Flow. Since the hosts are added when a Flow is created from a list of Flows/Jobs, during deserialization hosts were duplicated for each of the jobs contained in the Flow.

This PR should fix this.


utf:
Great, thank you.

PR #506, opened 2023-12-08
Add documentation for how to deal with stores beyond `MemoryStore`

Andrew-S-Rosen:
## Summary

This PR addresses a comment raised in https://github.com/openjournals/joss-reviews/issues/5995. We now include documentation for how to modify the Jobflow settings to use a custom data store. This was already described in the API documentation, but now a concrete example is provided in the main documentation.

One open question: There is basically the [same info](https://materialsproject.github.io/atomate2/user/install.html#jobflow-yaml) in a portion of the Atomate2 documentation. Do we instead just link to it to prevent duplication of content? Or is there value in having a dedicated page with Jobflow. I can see arguments for both. In any case, this PR provides an example.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -6,6 +6,7 @@
    Install jobflow <install>
    Install FireWorks (Optional) <install_fireworks>
    Tutorials <tutorials>
+   Configuring Data Stores <stores>
 
 .. toctree::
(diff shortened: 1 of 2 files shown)


utf:
Great, thank you.

PR #512, opened 2023-12-18
Optionally avoid deserialization when resolving references

gpetretto:
## Summary

Following the discussion about the need to deserialize objects when the output references are being resolved, I have added an option to avoid it. This would fail if the reference is a property of the deserialized object.


utf:
This looks great, thanks @gpetretto

PR #516, opened 2023-12-18
Fix handling of function from standard library

gpetretto:
## Summary

I was trying to use a function from the standard library to create a simple example and I got an error:

```python
import time
from jobflow import Job, SETTINGS

j = Job(time.sleep, (1,))
j.run(SETTINGS.JOB_STORE)
```

gives

> TypeError: time.sleep() takes exactly one argument (2 given)

The reason is that `sleep` has a `__self__` attribute and in `Job` it is then converted to a bound method. This seems to be shared by most (all?) the functions in the standard library. 

To adress this I changed the check performed in `Job.run()` assuming that if the `__self__` is a module, than the callable should not be converted to a method. I am not entirely sure if I may be missing some case where this is not true.

### DIFF
--- a/src/jobflow/core/job.py
+++ b/src/jobflow/core/job.py
@@ -554,7 +554,6 @@ class Job(MSONable):
         --------
         Response, .OutputReference
         """
-        import builtins
         import types
         from datetime import datetime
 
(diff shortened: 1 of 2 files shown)


utf:
Thanks @gpetretto

PR #528, opened 2024-01-17
CounterStore

jmmshn:
## Generical global counters might be useful for some special workflows

Many of the document models in emmet are designed to deal with either `int` or `"mp-"+int` type of ids for the different documents.  As such, they sometimes require unique integer counters for specific downstream documents to be built properly.
This feature should allow atomate2 to automatically assign unique ids to the tasks which (I believe) is currently unset.
If the task id is set I think more of the aggregated document model in emmet can be directly used in the middle of future more complex workflows.

An example of this problem is here:
https://github.com/materialsproject/atomate2/pull/655

Where the `InsertionElectrode` document expects task_id to be set.

### DIFF
--- a/src/jobflow/core/store.py
+++ b/src/jobflow/core/store.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import typing
 
 from maggma.core import Store
+from maggma.stores import MongoStore
 from monty.json import MSONable
 
 from jobflow.core.reference import OnMissing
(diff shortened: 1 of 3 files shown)


jmmshn:
> This feature should allow atomate2 to automatically assign unique ids to the tasks which (I believe) is currently unset.
If the task id is set I think more of the aggregated document model in emmet can be directly used in the middle of future more complex workflows.

@utf, I think a counter for the `TaskDocs` in atomate2 can be incorporated pretty easily by monkey patching the constructor method to call the counter incrementation.

utf:
Thanks @jmmshn, this is a nice implementation. As I mentioned on the atomate2 PR, the goal with jobflow is to move away from counters since they cause an issue when merging two databases. If MP need mp_id's then I suggest they be added in their ingestion pipeline. They shouldn't be needed for atomate2/jobflow.

jmmshn:
Closing this since we are resolving it in the other PR #529

Issue #519, opened 2024-01-01
Why UUIDs?


janosh:
I was looking at a wall of job IDs like this earlier and wondering if `jobflow` actually requires full 36-character UUIDs?

[...]

PR #529, opened 2024-01-18
allow different UID types

jmmshn:
## Only allow UUIDs that can be sorted in time

Partially addresses:
https://github.com/materialsproject/jobflow/issues/519

- Change all the function names  `*uuid` -> `*uid`
- Added new setting for `UID_TYPE`
- Added support for `uuid1` and `ulid` via `python-ulid`
- Added functions to retrieve time stamp information
- Added tests


gpetretto:
Thanks for implementing the change and making uuid4 the default. 
It would be convenient for me (and maybe for others), not to have an abrupt change in the exposed functions as I am using the `suuid` function to generate a string uuid in a few places. It is not such a big deal, but would it be fine to keep the `suuid` function in the `uuid` module with a deprecation warning?

jmmshn:
Yeah that makes sense.

jmmshn:
BTW does anyone know what the `s` in `suuid` stands for?

utf:
Thanks @jmmshn.

Is it possible to make `ulid` an optional dependency and only import it if the user has asked for it in their settings file. The code should also throw an appropriate error explaining which python package is needed to use `ulid` if it is not installed.

utf:
Thanks @jmmshn

jmmshn:
BTW, can we get a release with this, then I can finish the atomate2 PR.

utf:
Done (`v0.1.17`)

PR #544, opened 2024-02-14
jsanitize fireworks Task

gpetretto:
I have seen that in some cases the `JobFiretask` does not properly serialize a Job.
Consider the case of a Maker with a Callable as an attribute.

```python
from fireworks import LaunchPad
from jobflow.managers.fireworks import job_to_firework

from dataclasses import dataclass
from typing import Callable
from jobflow import Maker, job


def print_function():
    print("TEST")


@dataclass
class CallableMaker(Maker):
    f: Callable
    name: str = "CallableMaker"

    @job
    def make(self, *args, **kwargs):
        self.f()

j = CallableMaker(f=print_function).make()

fw = job_to_firework(j)
lpad = LaunchPad.auto_load()
lpad.add_wf(fw)
```

When running this, in the firework DB you have
```json
{
    "spec": {
        "_tasks": [
            {
                "job": {
                    "@module": "jobflow.core.job",
                    "@class": "Job",
                    "@version": "0.1.14.post122+g6439853",
                    "function": {
                        "@module": "__main__",
                        "@callable": "CallableMaker.make",
                        "@bound": {
                            "@module": "__main__",
                            "@class": "CallableMaker",
                            "@version": null,
                            "f": "<function print_function at 0x16ad82cb0>",
                            "name": "CallableMaker"
                        }
                    },
                   ...

}
```
Where the function is serialized as a string.

This PR addresses this issue by overriding the `JobFiretask.to_dict` method and explicitly jsanitizing the Job.

### DIFF
--- a/src/jobflow/managers/fireworks.py
+++ b/src/jobflow/managers/fireworks.py
@@ -5,6 +5,8 @@ from __future__ import annotations
 import typing
 
 from fireworks import FiretaskBase, Firework, FWAction, Workflow, explicit_serialize
+from fireworks.utilities.fw_serializers import recursive_serialize, serialize_fw
+from monty.json import jsanitize
 
 if typing.TYPE_CHECKING:
     from collections.abc import Sequence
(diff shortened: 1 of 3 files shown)


utf:
Great catch! Thanks @gpetretto

utf:
Is it possible to add a quick test?

gpetretto:
> Is it possible to add a quick test?

done

utf:
Perfect, thanks so much.

PR #559, opened 2024-03-04
Fix jsanitize in `contains_flow_or_job`

gpetretto:
This change in emmet https://github.com/materialsproject/emmet/commit/7735bb1a51cd9b339789bbdca0df92fbe366ddb8 leads to an error in `contains_flow_or_job`, due to the impossibility to call as_dict on some Enums. 
This PR adds the `enum_values=True` to handle those cases. 

In general, I am wondering if it is worth making this check every time a Flow is instantiated.

### DIFF
--- a/src/jobflow/utils/find.py
+++ b/src/jobflow/utils/find.py
@@ -205,7 +205,7 @@ def contains_flow_or_job(obj: Any) -> bool:
         # argument is a primitive, we won't find an flow or job here
         return False
 
-    obj = jsanitize(obj, strict=True, allow_bson=True)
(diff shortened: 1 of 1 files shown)


FabiPi3:
@gpetretto Fyi, I am trying to store the whole enum information in the output. I think this should fix this as well. Here are my pull requests: https://github.com/materialsvirtuallab/monty/pull/640 and https://github.com/materialsproject/jobflow/pull/565

What are your thoughts about this?

utf:
Is this PR still needed? It seems like compatability is now working?

PR #570, opened 2024-03-25
Add `job_dir` attribute to `Response` class to record where a job ran

janosh:
includes tests.

TODO

there are probably a bunch more places where the `job_dir` attribute should be set when executing a job.
also, should we add the same attribute to the `Job` class @utf? it would only be populated after the job has run which maybe makes for a worse UX?


utf:
Thanks @janosh. I think for now let's leave it off the `Job` object as really it is a runner feature.

PR #574, opened 2024-03-29
Migrate from `pip` to `uv` for CI dependency install

janosh:
`uv` continues to impress. 😄 

CI install time went from ~30s to 6s


utf:
Awesome, thanks!

PR #579, opened 2024-04-05
fix @job annotation (again)

FabiPi3:
This is a follow-up of #578 . As it turns out, `pycharm` only reacts on the `@overload` if the original function is also annotated. This is also in agreement with `mypy`.


utf:
Thanks!

PR #607, opened 2024-05-09
Add support for `pip install jobflow[remote]`

Andrew-S-Rosen:
I updated `pyproject.toml` to support doing `pip install jobflow[remote]` to install both `jobflow` and `jobflow-remote`. In principle, this will also allow `jobflow` to have some flexibility with setting version pins (if needed) on `jobflow-remote`.


Andrew-S-Rosen:
Eh, I guess this isn't really needed... I'll close this.

PR #640, opened 2024-07-04
Increase CI sensitivity by testing both lowest-direct and highest dependency resolution strategy

janosh:
related PRs: https://github.com/materialsproject/pymatgen/pull/3852 + https://github.com/CederGroupHub/chgnet/pull/159

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -35,14 +35,18 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.9', '3.10', '3.11']
+        version:
+          - { python: "3.9", resolution: highest, extras: strict }
+          - { python: "3.10", resolution: lowest-direct, extras: strict }
+          - { python: "3.11", resolution: highest, extras: non-strict }
+          - { python: "3.12", resolution: lowest-direct, extras: non-strict }
 
     steps:
       - uses: actions/checkout@v4
 
       - uses: actions/setup-python@v5
         with:
-          python-version: ${{ matrix.python-version }}
+          python-version: ${{ matrix.version.python }}
           cache: pip
           cache-dependency-path: pyproject.toml
 
@@ -52,13 +56,13 @@ jobs:
       - name: Install dependencies
         run: |
           pip install uv
-          uv pip install .[strict,tests] --system
+          uv pip install '.[tests,${{ matrix.version.extras }}]' --system --resolution=${{ matrix.version.resolution }}
 
       - name: Test
         run: pytest --cov=jobflow --cov-report=xml
 
       - uses: codecov/codecov-action@v1
-        if: matrix.python-version == '3.11' && github.repository == 'materialsproject/jobflow'
+        if: matrix.version.python == '3.11' && github.repository == 'materialsproject/jobflow'
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
           file: ./coverage.xml
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -65,6 +65,21 @@ strict = [
     "python-ulid==2.7.0",
     "typing-extensions==4.12.2",
 ]
+non-strict = [
+    "FireWorks>=2.0.3",
+    "PyYAML>=6.0.1",
+    "maggma>=0.69.0",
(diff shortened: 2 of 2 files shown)


janosh:
the py 3.12 CI run is failing due to a `pyzmq` build error

```
error: Failed to prepare distributions
  Caused by: Failed to fetch wheel: pyzmq==24.0.1
  Caused by: Failed to build: `pyzmq==24.0.1`
  Caused by: Build backend failed to build wheel through `build_wheel()` with exit status: 1
```

based on


```
$ pip show pyzmq              
Name: pyzmq
Version: 26.0.3
Location: /Users/janosh/.venv/py312/lib/python3.12/site-packages
Requires:
Required-by: ipykernel, jupyter-client, jupyter-server, maggma, notebook
```

looks like jobflow depends on `pyzmq` via `maggma` to increasing min `maggma` version to 0.69 which didn't help. @tschaume @tsmathis could we increase the min [`pyzmq` version here](https://github.com/materialsproject/maggma/blob/5693202219a13629e75cb482f692482414d13308/pyproject.toml#L41) to [25.1.1 which added 3.12 support](https://github.com/zeromq/pyzmq/blob/469d3075b45b5bc8578fc061c5e2f489d4fbfe16/docs/source/changelog.md#2511)

utf:
Hi @janosh, thanks for this. I'm a bit confused about what's happening here. If you use resolution strategy="highest" and the strict optional dependencies at the same time, won't this just use the pinned versions in strict? I think testing the pinned versions in strict is useful btw, but happy to add additional dependency tests on top of this.

tschaume:
@janosh I can't tell from your error message what exactly the build error is but if `pyzmq==25.1.1` still supports python 3.9, it should be ok to increase the minimum version in `maggma`. Do the jupyter packages that depend on `pyzmq` enforce a specific minimum version on pyzmq?

janosh:
> If you use resolution strategy="highest" and the strict optional dependencies at the same time, won't this just use the pinned versions in strict?

@utf that's right and hence be equivalent to CI as it runs prior to this PR. the 3-fold matrix strategy does the following

```yml
- { python: "3.9", resolution: highest, extras: "tests,strict" }  # this is equivalent to the previous CI run before this PR
- { python: "3.10", resolution: lowest-direct, extras: "tests" }  # test oldest allowed versions
- { python: "3.11", resolution: highest, extras: tests }  # test latest released versions
```

though i just noticed a typo in the middle case (which incorrectly used `strict`)

@tschaume doesn't look like it, definitely not [not ipykernel](https://github.com/ipython/ipykernel/blob/d4a8703593cbe5d82bc955d68543dc623f5b4510/pyproject.toml#L33), [nor notebook](https://github.com/jupyter/notebook/pull/7305). should i submit a PR or will you bump `pyzmq` in `maggma`?

janosh:
@utf looks like the new CI is working. it found `jobflow` is incompatible with the lowest `pydantic>=2.0.1` it listed, needs at least 2.4 😄

janosh:
only failing CI job now is code coverage which supposedly fell by 8%. makes no sense since no new code was added in this PR...

tschaume:
> should i submit a PR or will you bump `pyzmq` in `maggma`?

@janosh feel free to go ahead and submit a PR with maggma. @Andrew-S-Rosen and @rkingsbury might have some feedback and would be able to merge your PR.

janosh:
@utf i think this is ready to go except that the coverage seems to be thrown off by the use of `pytest.importorskip` on several optional dependencies which are only installed in the 1st of these 3 CI runs which uses the `strict` deps set. in the other 2 CI runs, tests requiring optional deps are skipped (i.e. tests that would fail if you just run `pip install jobflow` in an empty environment)

```yml
- { python: "3.9", resolution: highest, extras: "tests,strict" }  # this is equivalent to the previous CI run before this PR
- { python: "3.10", resolution: lowest-direct, extras: "tests" }  # test oldest allowed versions
- { python: "3.11", resolution: highest, extras: tests }  # test latest released versions
```

janosh:
@utf any chance you could update the required CI checks to match the new names including `lowest` / `highest` in the run name

utf:
Hi @janosh, I've updated the required CI, so this can be merged now.

Thanks very much for the addition.

Issue #639, opened 2024-07-03
BUG: append_name for Flows generated as a Response


JaGeo:
I am trying to implement a more complex workflow (see https://github.com/materialsproject/atomate2/pull/903). I need to renae jobs to work on the corresponding tests.

```Python
from jobflow import job, Flow, Response, run_locally


@job
def my_function(my_parameter: int) -> int:
    return my_parameter


@job
def my_second_function(my_parameter: int) -> int:
    job2 = my_function(my_parameter)
    return Response(replace=Flow(job2))


job3 = my_second_function(2)
job3.append_name(" test")
run_locally(job3)
```

This results in
```
2024-07-03 15:40:30,217 INFO Started executing jobs locally
2024-07-03 15:40:30,337 INFO Starting job - my_second_function test (bfde76e4-b636-416b-9fb9-ada2577f7377)
2024-07-03 15:40:30,339 INFO Finished job - my_second_function test (bfde76e4-b636-416b-9fb9-ada2577f7377)
2024-07-03 15:40:30,339 INFO Starting job - my_function (653a9993-4c4e-4332-9321-0f50ce985a55)
2024-07-03 15:40:30,340 INFO Finished job - my_function (653a9993-4c4e-4332-9321-0f50ce985a55)
2024-07-03 15:40:30,340 INFO Finished executing jobs locally
```

Is there another way to rename that I am overlooking? I would have hoped that "my_function" also gets renamed.

davidwaroquiers:
Hi @JaGeo 

Not sure if it is a bug but at least it is definitely a use case. Somehow what's missing is the dynamic application of the append procedure, similar to e.g. what is done with update_metadata (in which there is a dynamic argument, by default True).

I'll let @utf comment on this but it seems reasonable to add this feature.

utf:
I agree with @davidwaroquiers, we could add a `dynamic` option as we did here: https://github.com/materialsproject/jobflow/blob/11034dd00ad0ed64a027560a4f040c5b8c657193/src/jobflow/core/job.py#L935-L937

JaGeo:
@utf @davidwaroquiers adding a dynamic option would be great and would help with solving my issue.

davidwaroquiers:
I guess it would be good (and more consistent) to also add the same filtering options. Should then this stay within append_name or be a new update_name method?

PR #644, opened 2024-07-11
Add dynamic option to append_name

JaGeo:
Hi @utf , @davidwaroquiers ,

I have made a draft implementation to address #639 . Let me know what you think. I hope I haven't overlooked anything.

### DIFF
--- a/src/jobflow/core/flow.py
+++ b/src/jobflow/core/flow.py
@@ -601,7 +601,6 @@ class Flow(MSONable):
         for job in self:
             job.append_name(append_str, prepend=prepend, dynamic=dynamic)
 
-
     def update_metadata(
         self,
         update: dict[str, Any],
--- a/src/jobflow/core/flow.py
+++ b/src/jobflow/core/flow.py
@@ -582,7 +582,7 @@ class Flow(MSONable):
                 dict_mod=dict_mod,
             )
 
-    def append_name(self, append_str: str, prepend: bool = False):
+    def append_name(self, append_str: str, prepend: bool = False, dynamic: bool = True):
         """
         Append a string to the name of the flow and all jobs contained in it.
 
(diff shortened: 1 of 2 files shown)


JaGeo:
I am working on https://github.com/materialsproject/atomate2/pull/752. To finish it, this or a similar functionality would  be needed. Otherwise, `mock_vasp` does not work.

davidwaroquiers:
Hi @JaGeo 

I did not test but I think it should work. I guess for your own needs is fine but if I have to say one thing, maybe I would make a new update_name method that has more functionalities:

- Update the name without adding prefix/suffix (this is already available if you just do some_job.name = 'mynewname', but not in a dynamic way)
- Filters similar to update_metadata and others

API of this method could be:

```
def update_metadata(
        self,
        newname: str | None = None,
        prefix: str | None = None,
        suffix: str | None = None,
        name_filter: str = None,
        function_filter: Callable = None,
        dynamic: bool = True,
    ):
```

Then the append_name can just be used by this method if prefix and/or suffix is provided. (and the dynamic in append_name is not needed, I guess it would be better/less confusing if it is only in one place)

Not sure if it is absolutely essential, but it would make thing consistent with the rest of the update_* methods. (and I guess it would not be a huge work but again, it's all a matter of "is it needed and worth it if there is no use case right now ?")

JaGeo:
I don't really have a use case for the other cases. My use case currently to make sure I can differentiate jobs for moch_vasp.

(And, yes, it works. I tested it 😅)

utf:
Thanks, @JaGeo. I'll merge this for now. I put @davidwaroquiers comment into a new issue #648.

JaGeo:
Thanks, @utf . Would a new version be okay? Then, I can finalize the Grüneise workflow in atomate2 as well

utf:
I just pushed a new version (v0.1.18). It should be out soon.

JaGeo:
Thank you @utf !

PR #673, opened 2024-08-28
Fix docstring in Job.

davidwaroquiers:
The examples in the docstring of the Job class used args instead of function_args as an argument. These are fixed in this very small PR.


utf:
Thanks @davidwaroquiers

PR #679, opened 2024-09-18
Add `Flow.metadata` attribute and `Flow.update_metadata` method

janosh:
`Flow.update_metadata` is now much more similar to and has reached feature parity with `Job.update_metadata`:

https://github.com/materialsproject/jobflow/blob/367836b0d51a89fa58a9d685c4cdb9fc7f806c01/src/jobflow/core/job.py#L922-L930

### DIFF
--- a/src/jobflow/core/flow.py
+++ b/src/jobflow/core/flow.py
@@ -659,13 +659,13 @@ class Flow(MSONable):
         """
         from jobflow.utils.dict_mods import apply_mod
 
-        if target in ["flow", "both"] and name_filter in (None, self.name):
+        if target in ("flow", "both") and name_filter in (None, self.name):
             if dict_mod:
                 apply_mod(update, self.metadata)
             else:
                 self.metadata.update(update)
 
-        if target in ["jobs", "both"]:
+        if target in ("jobs", "both"):
             for job in self:
                 job.update_metadata(
                     update,
@@ -675,7 +675,7 @@ class Flow(MSONable):
                     dynamic=dynamic,
                 )
 
-        if dynamic and target in ["jobs", "both"]:
+        if dynamic and target in ("jobs", "both"):
             dict_input = {
                 "update": update,
                 "name_filter": name_filter,
--- a/src/jobflow/core/flow.py
+++ b/src/jobflow/core/flow.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 import logging
 import warnings
 from copy import deepcopy
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Literal
 
 from monty.json import MSONable
 
@@ -128,6 +128,8 @@ class Flow(MSONable):
         order: JobOrder = JobOrder.AUTO,
         uuid: str = None,
         hosts: list[str] = None,
+        metadata: dict[str, Any] = None,
+        metadata_updates: list[dict[str, Any]] = None,
     ):
         from jobflow.core.job import Job
 
@@ -141,6 +143,8 @@ class Flow(MSONable):
         self.order = order
         self.uuid = uuid
         self.hosts = hosts or []
+        self.metadata = metadata or {}
+        self.metadata_updates = metadata_updates or []
 
         self._jobs: tuple[Flow | Job, ...] = ()
         self.add_jobs(jobs)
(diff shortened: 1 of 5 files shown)


janosh:
@utf i'm sure you're strapped for time but even partial feedback here would be much appreciated!

src/jobflow/core/flow.py
```diff
@@ -647,14 +657,33 @@ def update_metadata(
...
+        from jobflow.utils.dict_mods import apply_mod
+
+        if target in ["flow", "both"] and name_filter in (None, self.name):
+            if dict_mod:
+                apply_mod(update, self.metadata)
+            else:
+                self.metadata.update(update)
+
+        if target in ["jobs", "both"]:
+            for job in self:
```
utf, line 669: Here `job` could actually be a nested flow. So you should still iterate over these if `target = "flow"`.

utf (commented):
Thanks @janosh. My only concern is about breaking the API of `update_metadata` between `Flow`s and `Job`s. If we add the `target` option to `Job.update_metadata` that should make this function easier to use and the implementation cleaner.

src/jobflow/core/flow.py
janosh: perhaps the naming could be improved. `target = "flow"` isn't meant to update the metadata of all nested flows (to the exclusion of jobs) but rather only the flow itself, not any of its jobs or nested flows.

how about we rename from `target: Literal["flow", "jobs", "both"] = "both"` to `target: Literal["self", "nested", "both"] = "both"`? unless you think it's important to have more control, i.e. be able to only update nested `Flow` or `Job` metadata. there might be use cases for that. in which case maybe we want `target: Literal["self", "nested", "jobs", "flows", "both"] = "both"`?

janosh: to clarify this could be the doc string:

```
        target
            Specifies where to apply the metadata update. Options are:

            - "self": Update only the Flow's own metadata
            - "nested": Update only the metadata of Jobs/Flows within the Flow
            - "jobs": Update only the metadata of Jobs within the Flow
            - "flows": Update only the metadata of Flows within the Flow
            - "all": Update both the Flow's metadata and nested Job+Flow metadata (default)
```

utf: I think this is overcomplicating it. If you want to select specific flows or jobs then you can pass in a name or class filter. So no need for the extra options. The main thing is the API should be consistent between jobs/flows.

janosh: makes sense. then the easiest thing would be to get rid of `target` altogether and use the `"all"` behavior?

janosh: > I think the benefit of target is that it would be a shortcut for `class=Flow` vs `class=Job`. I would be happy either having the target option or specifying the shortcut in the docstring.

i don't follow. by `class=Job|Flow`, did you mean `filter_function=Job|Flow`? the job variant `filter_function=Job` doesn't seem to work the way you're suggesting so maybe you meant sth else?

utf: Yes, I was thinking of `class_filter`. I think the `callback_filter` you proposed sounds very flexible and would be useful!

utf:
@janosh I lost track of this. Is it good to go?

janosh:
@utf i think this is ready to go now. a future PR could add the `callback_filter` keyword to other update methods like `update_kwargs` as well

janosh:
hmmm... added a lot of test cases. not sure what else to add. could it be the coverage calculation is off?

utf (commented):
@janosh just one comment. Happy for you to merge once that's done. I can then push a new version.

src/jobflow/core/job.py
```diff
@@ -927,6 +926,7 @@ def update_metadata(
         function_filter: Callable = None,
         dict_mod: bool = False,
         dynamic: bool = True,
+        callback_filter: Callable[[jobflow.Flow | Job], bool] = lambda _: True,
```
esoteric-ephemera, line 929: @janosh: the `Callable` `callback_filter`s seem to break some functionality, getting `'dict' object is not callable` when these are decoded. Double checked manually that `MontyDecoder` is not rehydrating these into callables, just as `dict`s

janosh: Or you issue an error when someone tries to serialize a callback_filter to explain that this feature is only compatible with create-and-run kind of workflows

utf: Serialisation shouldn't be a problem if the function is importable, e.g if it is defined in atomate2 or another package.

esoteric-ephemera: The default `callback_filter` isn't serializing correctly - I'll look into it a bit more and update you both

PR #721, opened 2024-12-17
Change default callback_filter in update_metadata

esoteric-ephemera:
Changing the default value of `callback_filter` in the `update_metadata` function of both `Job` and `Flow` from a lambda to `None`. Having its default value be a lamdba function means that it cannot be round-trip de-/serialized:
```
from monty.json import MontyEncoder, MontyDecoder

f = lambda _ : True
encoded = MontyEncoder().encode(f)
print(encoded)
>>> '{"@module": "__main__", "@callable": "<lambda>", "@bound": null}'
decoded = MontyDecoder().decode(encoded)
print(decoded)
>>> {'@module': '__main__', '@callable': '<lambda>', '@bound': None}
```
This causes issues in fireworks with the newest release of jobflow. Linking relevant discussion [here](https://github.com/materialsproject/jobflow/pull/679#discussion_r1866290870) and tagging @janosh  and @utf for thoughts

Because lambda's aren't pickle-able without something like dill, writing a `callback_filter` as an importable function might be the only option for fireworks / jobflow users

### DIFF
--- a/src/jobflow/core/flow.py
+++ b/src/jobflow/core/flow.py
@@ -617,7 +617,7 @@ class Flow(MSONable):
         function_filter: Callable = None,
         dict_mod: bool = False,
         dynamic: bool = True,
-        callback_filter: Callable[[Flow | Job], bool] = lambda _: True,
+        callback_filter: Callable[[Flow | Job], bool] | None = None,
     ):
         """
         Update the metadata of the Flow and/or its Jobs.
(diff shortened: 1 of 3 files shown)


janosh (approved):
Looks good 👍  thank you for fixing!

utf (approved):
Thanks @esoteric-ephemera

PR #815, opened 2025-10-23
A flow decorator

vineetbansal:
## Summary

I was talking to @Andrew-S-Rosen about implementing a `@flow` decorator for `jobflow`, and a previous [discussion](https://github.com/materialsproject/jobflow/discussions/416) on this topic. Here's an initial stab at it. It's using a "flow context" (an idea also used in the Prefect library) so that a job can figure out if it's running inside a decorated flow function, and if so, add itself to the flow's job list. I believe this should help us get around the "updating global variables" issue as mentioned in a related [PR](https://github.com/materialsproject/jobflow/pull/547) in the linked discussion.

- Added a `@flow` decorator
- Added some tests, one for a dynamic run with a `replace`ment.

None

## TODO (if any)

I have almost no history with using `jobflow`, so certainly I'm not thinking about all use cases, so any pointers (initially failing tests that I can add to cover different scenarios before tackling them) will be welcome.

## Checklist

title.

  type check your code.

prior to allowing commits.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,28 +41,28 @@ dependencies = [
 ulid = ["python-ulid"]
 docs = [
     "autodoc_pydantic==2.2.0",
-    "furo==2025.9.25",
-    "ipython==9.3.0",
+    "furo==2025.12.19",
+    "ipython==9.9.0",
     "myst_parser==4.0.1",
     "nbsphinx==0.9.8",
     "sphinx-copybutton==0.5.2",
     "sphinx==8.1.3",
 ]
 dev = ["pre-commit>=2.12.1", "typing_extensions; python_version < '3.11'"]
-tests = ["moto==5.1.18", "pytest-cov==7.0.0", "pytest==9.0.2"]
+tests = ["moto==5.1.20", "pytest-cov==7.0.0", "pytest==9.0.2"]
 vis = ["matplotlib", "pydot"]
 fireworks = ["FireWorks"]
 strict = [
     "FireWorks==2.0.8",
     "PyYAML==6.0.3",
     "maggma==0.72.0",
-    "matplotlib==3.10.7",
+    "matplotlib==3.10.8",
     "monty==2025.3.3",
-    "moto==5.1.18",
+    "moto==5.1.20",
     "networkx==3.4.2",
     "pydantic-settings==2.12.0",
     "pydantic==2.12.5",
-    "pydash==8.0.5",
+    "pydash==8.0.6",
     "pydot==4.0.1",
     "python-ulid==3.1.0",
     "typing-extensions==4.15.0",


vineetbansal:
@Andrew-S-Rosen, @davidwaroquiers , @utf

vineetbansal:
At a high level, to break down what's happening:

- The `flow` decorator sets a context variable, and returns a `DecoratedFlow`, a subclass of `Flow`, which saves the callable, args and kwargs.
- A `Job`, when run, checks to see if its running in the context of a `flow` decorator, if so, it adds itself to the `DecoratedFlow`.
- `run_locally` runs the `DecoratedFlow` by calling it with its saved args/kwargs, and at the end, resolves any references in the output.

A parallel change would have to be made in `jobflow-remote` if this is on the right path.

I've added some more tests to cover what I think would be typical scenarios in using the `@flow` decorator:

```

@flow
def my_flow(a, b):
  # typical case
  return add(a, b).result  # or the inadvertent shorthand `return add(a, b)` which is handled by existing code correctly.

@flow
def my_flow():
   # some jobs are run, but only for their side effects
   return 42
   
@flow
def my_flow(a, b):
  # some complex structure with hashable keys but jobs/flows as values     
  return [add(a, b), add(b, a).output], {
      "foo": add(a, b).output,
      "bar": [add(a, b), subtract(a, b)],
  }

# replaced job in a flow
@job
def replace_job(x, y):
    another_job = subtract(x, y)
    return Response(replace=another_job)
    
@flow
def my_flow(a, b):
  return add(a, b).result
```

gpetretto:
Hi @vineetbansal and @Andrew-S-Rosen,
thanks a lot for implementing this. I have made some tests and here are my initial comments. 
* It is not clear why for a `DecoratedFlow` the output should be different from the standard case of a normal Flow. In the example below I would expect that the output of the two executions would be equivalent.
  ```python
  @flow
  def my_flow(a, b):
      j= add(a, b)
      return j
  
  j = add(3,4)
  f = Flow(j, output=j.output)
  
  result = run_locally(f, ensure_success=True)
  result = run_locally(my_flow(3,4), ensure_success=True)
  ```
  This, however, is just concerning how `run_locally` generates the outputs 
* Just to clarify, I suppose there is no aim at executing any piece of code inside a `flow` decorated function, aside from combining jobs actions. For example the following code raises an error:
  ```python
  @flow
  def my_flow(a, b):
      j1= add(a, b)
      x = j1.output + 5
      j2 = add(x, b)
      return j2.output
  ```
  I would not see any reasonable way of implementing this, especially for jobflow-remote or fireworks executions. Just for my undestanding, is something similar possible in prefect?
* It seems that is not possible to nest multiple `flow` decorated functions. The following code does not execute any Job and returns None:
  ```python
  @flow
  def add_single(a, b):
      j1 = add(a, b)
      return add(j1.output, 2)
  
  @flow
  def add_combine(a, b):
      f1 = add_single(a, b)
      return add_single(f1.output, 3)
  
  f = add_combine(1, 2)
  result = run_locally(f, ensure_success=True)
  ```
  Maybe here it is possible to change the `flow` decorator so that when the function is called can also check if it is inside the `flow_build_context` and trigger a recursive unravel of the flow functions? In general I would expect this to work more or less like the following code based on Makers:
  ```python
  class AddSingleMaker(Maker):
      name = "single"
  
      def make(self, a, b):
          j1 = add(a, b)
          j2 = add(j1.output, 2)
          return Flow([j1, j2], output=j2.output)
  
  class AddCombineMaker(Maker):
      name = "combine"
  
      def make(self, a, b):
          f1 = AddSingleMaker().make(1, 2)
          f2 = AddSingleMaker().make(f1.output, 3)
          return Flow([f1, f2], output=f2.output)
  
  f = AddCombineMaker().make(1, 2)
  result = run_locally(f, ensure_success=True)
  ```
* Concerning jobflow-remote, as things are now in the end I now think it would be relatively easy to integrate this. My understanding is that the `flow` decorated function should just be used to first generate the "complete" Flow object. I have tried the code below in jobflow-remote and it seems to be working as expected
  ```python
  @flow
  def my_flow(a,b):
      j= add(a,b)
      return j
  
  print("create flow")
  f = my_flow(1, 2)
  
  with flow_build_context(f):
      output = f.fn(*f.args, **f.kwargs)
  
  f.name = f.fn.__name__
  f.output = output.output
  
  submit_flow(f, worker="local_shell")
  ```
  So I can basically include this check below in the `submit_flow` function and I would expect this to work:
  ```python
  if isinstance(flow, DecoratedFlow):
    with flow_build_context(flow):
        output = flow.fn(*flow.args, **flow.kwargs)
    
    flow.name = flow.fn.__name__
    flow.output = output.output
  ```
  Maybe the setting of `name` and `output` can even be done directly by the `flow` decorator?
  If instead you expect that the code will evolve to include code execution as in the example above or other functionalities things may get more complicated.
* If this solution works, I would expect that it may be applied in the same way for `flow_to_workflow` for Fireworks.

vineetbansal:
Thanks @gpetretto - your comments were most insightful and after a few iterations of overly complicating and then simplifying things on my end, I think I've ended up at more or less the same place as you had intuited.

> It is not clear why for a DecoratedFlow the output should be different from the standard case of a normal Flow..

Indeed - my original intuition was that the `@flow` decorator could be used on arbitrary functions, not necessarily ones for accumulating `Job`s and `Flow`s, so I was trying to resolve the output (to an admittedly incomplete extent) before returning it. I see that all the resolving is happening in the `responses` dict anyway, so that simplifies things on my end (and gets rid of a few test cases related to this return behavior).

> Just to clarify, I suppose there is no aim at executing any piece of code inside a flow decorated function, aside from combining jobs actions. .. I would not see any reasonable way of implementing this, especially for jobflow-remote or fireworks executions.

Yes - I talked to @Andrew-S-Rosen and we're not using `@flow` decorators (across workflow libraries) to execute code other than as a glue to create a DAG. So my initial (incomplete) attempt at trying to do this is not needed.

> is something similar possible in prefect?

Yes - it seems to be. Prefect seems to be passing in the "manager" to the context to execute arbitrary code in the flow:
```
    with engine.start():
        while engine.is_running():
            with engine.run_context():
                engine.call_flow_fn()
 ```
 
> It seems that is not possible to nest multiple flow decorated functions. .. Maybe here it is possible to change the flow decorator so that when the function is called can also check if it is inside the flow_build_context and trigger a recursive unravel of the flow functions?

This should now be fixed. If I understand correctly, the current `DecoratedFlow` should add itself to the context before "executing" (building the DAG for) the decorated function, and the `DecoratedFlow` should check if it is nested inside another (by checking the flow context) to add itself to the parent flow, which is what I'm doing now.

I've added a test case for this.

> Maybe the setting of name and output can even be done directly by the flow decorator?

Good idea. Done!

As it stands now, `Response(replace=..)` and dynamic flows are not currently working with this approach, and represented by two `xfail` tests.

This seems to have to do with the creation of brand-new `Flow` objects, independent of existing flows, when replace is called:
https://github.com/materialsproject/jobflow/blob/d208e460d7bad3533a6323703276cf75304c0891/src/jobflow/core/job.py#L1383

and

https://github.com/materialsproject/jobflow/blob/d208e460d7bad3533a6323703276cf75304c0891/src/jobflow/core/job.py#L1409

These 2 are the cases that we'd like to support in jobflow to be able to use it in our project. I'll dig into this some more, unless you have some thoughts already. I didn't want to delay this reply any more!

**Comment by gpetretto** [comment]
Hi @vineetbansal, thanks a lot for the updates! I will have a look at them. I just wanted to quickly clarify one point: I tried to fetch your latest changes and run the tests in `test_flow_decorator.py`, but they all seem to pass on my laptop. Even those marked with xfail. Could there be some differences depending on the environment? Or am I missing something?

**Comment by vineetbansal** [comment]
@gpetretto - no - you're right. Somewhere along the line during my refactoring the `Response(replace=..)` seems to have been fixed - I should have used `strict=True` for those xfails. I'm adding another commit that removes those xfails. Fingers crossed..

**Comment by vineetbansal** [comment]
@gpetretto - when you're looking at this PR, I wanted to bring something else to your attention as well, as I was going through its possible integration with [quacc](https://quantum-accelerators.github.io/quacc/user/wflow_engine/wflow_engines2.html#__tabbed_1_6). This is again going back to your question:

> It is not clear why for a DecoratedFlow the output should be different from the standard case of a normal Flow..

In the case of classic flows, the caller who creates the `Flow` is responsible for reaching into the results depending on what they "know" the eventual output of interest is (presumably by how they put the jobs together):
```
  job1 = add(1, 2)
  job2 = mult(job1.output, 3)
  flow = jf.Flow([job1, job2])

  responses = run_locally(flow, ensure_success=True)
  assert responses[job2.uuid][1].output == 9
```
or, the same thing but during `Flow` construction time:
```
  flow = jf.Flow([job1, job2], output=job2.output)

  responses = run_locally(flow, ensure_success=True)
  assert responses[flow.output.uuid][1].output == 9
```

However, in the case of a `@flow` decorated function (which should be opaque to the caller except for the input and output), it's not clear what uuid in the results the caller should be reaching in for (since the jobs/flows are all internal details of the `@flow`). For it to "just work", the use case would look like:
```
  @flow
  def workflow():
      job1 = add(1, 2)
      job2 = mult(job1.output, 3)
      return job2.output  # or `return job`

  response = run_locally(workflow(), ensure_success=True)
  assert response == 9
 ```

The latter is the use case where we use `jobflow` in quacc recipes.

We have a small tweak to this PR [here](https://github.com/vineetbansal/jobflow/pull/2/files) that allows us to do this. Perhaps we can incorporate something like that into this PR (either now or as a separate PR?)

The way I envision it working would be (in case you do want to preserve compatibility with `Flow`):
```
@flow
def workflow(...)
   # return a dict of responses, keyed by uuid, similar to Flow()
   
@flow(return_dict=False)
def workflow(...)
  # return the resolved output
```

This also allows us to support cases where the output is not a single `OutputReference` or `Flow`, but could easily be resolved anyway:
```
  @flow(return_dict=False)
  def my_flow(a, b):
      return [add(a, a), add(b, b), add(a, b)]
```

Andrew-S-Rosen:
Just adding a note for @vineetbansal that the `Maker` class can be found in `jobflow.core.maker.Maker`. Additional context can be found [here](https://materialsproject.github.io/jobflow/tutorials/6-makers.html).

vineetbansal:
@gpetretto - thanks for the detailed feedback and the fixes for the more involved tests. Yes - your fix regarding having the context as a list instead of a single `Flow` and only adding jobs that have their `host` as `None` makes sense, and still allows all our tests in our `quacc` repo to pass (I tested against this branch).

Thanks also for the fix for the `Maker`s.

For the one minor thing that you're flexible on:

> However, I do not have a strong opinion on this matter. If the replacing of the output is left as is, I would at least suggest to explicitly mention in the warning message that the output has been replaced.

We're able to handle more complicated cases of `@flow` functions returning list/dicts etc. in our code, so it doesn't necessarily have to go in the wrapper. However, inclusion of the following snippet as I've done in my latest commit to this branch (after merging your changes), and being explicit about the automatic replacement:

```
        if isinstance(output, (jobflow.Job, jobflow.Flow)):
            warnings.warn(
                f"@flow decorated function '{name}' contains a Flow or"
                f"Job as an output. Usually the output should be the output of"
                f"a Job or another Flow (e.g. job.output). Replacing the"
                f"output of the @flow with the output of the Flow/Job."
                f"If this message is unexpected then double check the outputs"
                f"of your @flow decorated function.",
                stacklevel=2,
            )
            output = output.output
```

This will allow our "recipes" to work seamlessly with `jobflow` in addition to `prefect`/`parsl` etc, (because the latter workflow engines make no distinction between a job and its output), so hopefully this is still okay.

If you're satisfied with these changes, then perhaps we prod @utf  to merge this? I'd be happy to add any additional docstrings/testcases of course, but otherwise I think we're there now..

gpetretto:
Hi @vineetbansal, thanks for checking the changes and including them in the PR.

> The intention is simply to construct the `Job`s that go inside the `@flow`, and the `[add(a, a), add(b, b)]` part (or that anything is returned really) is just an excuse to construct the appropriate `Job`s. We're able to handle this (and more complicated cases of `@flow` functions returning list/dicts etc). in our code, so it doesn't have to go in the `jobflow`.

I just have a doubt about this point. Since the resolution of the references is done inside `jobflow`, how do you deal with its inconsistencies inside quacc, if that is sitting on top of jobflow? Do you actually change the Flow output there?

If you would rather keep the replacement of the output, I wonder if it may be worth to recursively explore the output object (similarly to what is done [here](https://github.com/materialsproject/jobflow/blob/main/src/jobflow/utils/find.py#L181))
and replace all the instances of `Job` and `Flow` with their outputs.

> If you're satisfied with these changes, then perhaps we prod @utf to merge this? I'd be happy to add any additional docstrings/testcases of course, but otherwise I think we're there now..

I don't have further comments and I think the code is at a good stage. I agree that this can probably be merged, so @utf can review the changes. 
If everything is fine with him, maybe a small section in the documentation can be added, so that other users can discover the new functionality?

gpetretto:
Hi @Andrew-S-Rosen,

to be fair I would say that I find rather confusing the notation
```python
@flow
@job
def dynamic_job(maker):
```
At first sight I do not believe its meaning is obvious. I am also not sure how well this fits in the current `@flow` implementation. 
Wouldn't this change be more related to the `@job` decorator rather than to `@flow`? Maybe with an argument, like `@job(replace=True)`, or with a specific decorator `@replace` to combine with `@job`

```python
@replace
@job
def dynamic_job(maker):
```
I think this may have some benefits: 
* similar options could be also used to handle `addition` and `detour`
* It will not mess with the implementation of `@flow`, so it should be easier to implement
* at least in my opinion, the effect of such options would be clearer.

There is also a completely different alternative. Right now if a Job returns an instance of `Job` or `Flow`, the code prints a warning, but leaves things as they are, using the Job object as output (which I cannot think of a use case for). We could change the current implementation so that if the output is an instance of a `Job` or `Flow` the code will automatically interpret this as a replace. This would not even require additional objects or notations. The downside is that this would be a backward incompatible change, although it seems unlikely that anyone would have a `Flow` returned from their jobs now.

vineetbansal:
Thanks @gpetretto , @Andrew-S-Rosen. I played around with @gpetretto 's idea of "..could change the current implementation so that if the output is an instance of a Job or Flow the code will automatically interpret this as a replace." (+support for list of `Job`s which is our common use-case), by adding a bit of logic at the top of `Response.from_job_returns` method:

```
--- a/src/jobflow/core/job.py
+++ b/src/jobflow/core/job.py
@@ -1286,6 +1286,12 @@ class Response(typing.Generic[T]):
         Response
             The job response controlling the data to store and flow execution options.
         """
+        should_replace = isinstance(job_returns, Job)
+        if isinstance(job_returns, (list, tuple)):
+            should_replace = all(isinstance(resp, Job) for resp in job_returns)
+        if should_replace:
+            job_returns = Response(replace=job_returns)
+
         if isinstance(job_returns, Response):
             if job_returns.replace is None:
                 # only apply output schema if there is no replace.
```
and tried it on:
```
@job
def add(x ,y):
    return x + y

@job
def make_list_of_3(a):
    return [a] * 3

@job
def add_distributed(list_a):
    jobs = [add(val, 1) for val in list_a]
    return jobs
```

and its working as expected with `run_locally` as well as jobflow-remote's `submit_flow` (both with `scheduler_type: shell` and `scheduler_type: slurm`).

If this looks promising, perhaps we can open a new PR on this, since it is unrelated to the `@flow` decorator? I'm sure there are edge cases I may not be thinking about.

utf:
Firstly, appologies for the long delay getting to this PR. 

Thanks very much @vineetbansal for this contribution and for the discussion from @gpetretto and @Andrew-S-Rosen. The implementation looks surprisingly clean. I'm glad you were able to implement it with so few changes.

I'm happy to merge this. We can look at updating the documentation to highlight this new feature.

PR #823, opened 2025-11-21
getitem for Job delegates to its output

vineetbansal:
## Summary

This is essentially trying to resurrect the idea of `__getitem__` access to a `Job` returning the item from its `.output`, as in a [previous PR](https://github.com/materialsproject/jobflow/pull/450), but limited in scope to avoid breaking changes.

A use-case for this (included here as a test) is:
```
    @job
    def make_str(s):
        return {"hello": s}

    @job
    def capitalize(s):
        return s.upper()

    job1 = make_str("world")
    job2 = capitalize(job1["hello"])  # <---- instead of `job1.output["hello"]`

    flow = Flow([job1, job2])
```

### Context

While working with our package that uses `jobflow`, `prefect`, `parsl` and other workflow engines under the hood, we've recognized that flows are expressed succinctly as lookups on jobs:
```
    @job
    def greetings(s):
        return {"hello": f"Hello {s}", "bye": f"Goodbye {s}"}

    @job
    def upper(s):
        return s.upper()

    @flow
    def greet(s):
        job1 = greetings(s)
        job2 = upper(job1["hello"])
        return job2
 ```
 
While the `@flow` decorator is a [parallel effort](https://github.com/materialsproject/jobflow/pull/815) that we hope can be merged in soon, this PR focuses only on the `getitem` part.

The [existing PR](https://github.com/materialsproject/jobflow/pull/450) on this from a couple of years ago was also trying to integrate attribute access to jobs and flows, which is a more overreaching change, and calls for more internal refactoring than what we feel might be needed. For example, one suggested change there:
```
def __getattr__(self, name: str) -> OutputReference:
    if attr := getattr(self.output, name, None):
        return attr
    raise AttributeError(f"{type(self).__name__} has no attribute {name!r}")
```

would necessitate that all current and future attributes of `Job`s and `OutputReference`s be non-overlapping, which might be more trouble for development than is worth. A dict-lookup should not have the same disruptive effect from the point of view of the codebase, but will provide a useful shortcut for flow creation.

## Checklist

  type check your code.

prior to allowing commits.

### DIFF
--- a/src/jobflow/core/job.py
+++ b/src/jobflow/core/job.py
@@ -389,6 +389,24 @@ class Job(MSONable):
         name, uuid = self.name, self.uuid
         return f"Job({name=}, {uuid=})"
 
+    def __getitem__(self, key: Any) -> OutputReference:
+        """
+        Get the corresponding `OutputReference` for the `Job`.
+
+        This is for when it is indexed like a dictionary or list.
+
+        Parameters
+        ----------
+        key
+            The index/key.
+
+        Returns
+        -------
+        OutputReference
+            The equivalent of `Job.output[k]`
+        """
+        return self.output[key]
+
     def __contains__(self, item: Hashable) -> bool:
         """
         Check if the job contains a reference to a given UUID.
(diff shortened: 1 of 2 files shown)


utf:
I'm confused why this PR says merged when I didn't merge it, and the changes are not present in the jobflow repo. I'm going to try "reverting" the PR (nothing will change), and reopening this PR.

PR #839, opened 2026-01-15
Validate properties attribute in output_schema

gpetretto:
As discussed in https://github.com/materialsproject/atomate2/issues/1353 and https://github.com/materialsproject/atomate2/pull/1354, currently a code like 

```python
from pydantic import BaseModel
from jobflow import job, Flow, run_locally


class M(BaseModel):
    a: int

    @property
    def a2(self):
        return self.a * 2    


@job(output_schema=M)
def test():
    return M(a=2)

@job
def print_job(x):
    print(f"{x}")

j1 = test()
j2 = print_job(j1.output.a2)

f = Flow([j1,j2])

run_locally(f)
```

Fails with an error:
> AttributeError: M does not have attribute 'a2'.

but if the `(output_schema=M)` is removed from the job definition the code is executed correctly, meaning that it is just the validation step that fails to identify the existence of the properties. 

This PR aims at making the code above correct even when the schema is enforced, by checking the existence of property-like attributes (e.g. `@property`, functools' `@cached_property`, monty's `@lazy_property`) matching the name in the output schema. 

The current implementation does **not** attempt to extract the output type of the object returned by the property to verify if it is itself a subclass of `BaseModel`. I tried playing a bit with extracting information from the typing of the method, but it seems hard to cover all the possible cases and is probably a less interesting case to cover. It should be easier to cover at least the standard `@property`, if this is seen as necessary.
Another option would be to implement a `@typed_property(output_type=....)` decorator, for the cases where it may be important to enforce the checks in the output model.

### DIFF
--- a/src/jobflow/core/reference.py
+++ b/src/jobflow/core/reference.py
@@ -514,12 +514,47 @@ def validate_schema_access(
         The BaseModel class associated with the item, if any.
     """
     schema_dict = schema.model_json_schema()
-    if item not in schema_dict["properties"]:
+    item_in_schema = item in schema_dict["properties"]
+    property_like = isinstance(item, str) and has_property_like(schema, item)
+    if not item_in_schema and not property_like:
         raise AttributeError(f"{schema.__name__} does not have attribute '{item}'.")
 
     subschema = None
-    item_type = schema.model_fields[item].annotation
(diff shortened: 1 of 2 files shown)


utf:
Thanks @gpetretto. Handling the schema validation in an exhaustive way is quite tricky, but this looks like a nice extension.

Issue #855, opened 2026-03-03
BUG: job that returns a numpy array fails


FabiPi3:
Consider this example:

```python
import numpy as np
from jobflow import job, run_locally

@job
def return_numpy():
    return np.array([[1, 2], [3, 4]])

run_locally(return_numpy(), ensure_success=True)
```

While I expect this to work, the output can literally be anything, it actually fails. Error message:

```bash
Traceback (most recent call last):
  File "/Users/fabianpeschel/miniconda3/envs/exwoflow313/lib/python3.13/site-packages/jobflow/managers/local.py", line 117, in _run_job
    response = job.run(store=store)
  File "/Users/fabianpeschel/miniconda3/envs/exwoflow313/lib/python3.13/site-packages/jobflow/core/job.py", line 630, in run
    response = Response.from_job_returns(
        response, self.output_schema, job_dir=job_dir
    )
  File "/Users/fabianpeschel/miniconda3/envs/exwoflow313/lib/python3.13/site-packages/jobflow/core/job.py", line 1315, in from_job_returns
    if job_returns and isinstance(job_returns, (list, tuple)):
       ^^^^^^^^^^^
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

The bug appeared when switching to the newest version 0.3.1. I think the issue was introduced in here: https://github.com/materialsproject/jobflow/pull/848
The test `if job_returns` doesn't work for arrays. One would need a separate handling for this.

Probably @gpetretto and @utf can have a look.

davidwaroquiers:
Thanks @FabiPi3 


Truthiness in python at its best ... Sometimes I wonder why this is a python feature.


It is indeed a bug (I think, @gpetretto can you confirm). I think the fix is easy, is to replace:

`if job_returns and isinstance(job_returns, (list, tuple)):`

by 

`if job_returns is not None and isinstance(job_returns, (list, tuple)):`

@gpetretto What do you think ?

davidwaroquiers:
Checking a bit more I think it might need to be something like:

`if isinstance(job_returns, (list, tuple)) and len(job_returns) > 0:`

In my previous suggestion, an empty list again gives a replace because all([]) is True..

gpetretto:
I think this last option makes sense. I proposed to fix this in #856, so that it also takes into account the `dict` case.

PR #856, opened 2026-03-04
A return type of a dict of jobs from a Job is interpreted as a replace

vineetbansal:
## Summary

In a [previously accepted PR](https://github.com/materialsproject/jobflow/pull/833), we put in some logic to check if a `Job` (or a tuple/list of `Job`s) is returned by a `@job`. In those cases it is automatically interpreted as a `Replace`.

The motivation for that PR was that the `@job` specifies the dependencies, but the actual mechanics of replacement (Response/replace) are handled internally in `jobflow`. This has worked well for our suite of recipes, but we've stumbled on a use-case that might need one small enhancement to this behavior.

We found that since a `@job` that spawns other `Job`s needs to return *all* of them instead of the ones that may be of direct interest to the caller (otherwise the non-returned jobs are never added to the DAG), we're finding that it is useful in certain jobs to *namespace* the dependent `Job`s for easier use by the caller by returning a dict. Something like:
```

@job
def add
..

@job
def sub
..

@job
def add_distributed(list_a):
    add_jobs = [add(val, 1) for val in list_a]
    sub_jobs = [sub(val, 1) for val in list_a]
    return {"add": add_jobs, "sub": sub_jobs}
```

This PR adds this functionality while preserving the output structure. I've included a test to demonstrate this.
This change allows `jobflow` to work with all our recipes.

Let me know if there are any questions or concerns with this. Thanks!

## Checklist

title.

  type check your code.

prior to allowing commits.

### DIFF
--- a/src/jobflow/core/job.py
+++ b/src/jobflow/core/job.py
@@ -1265,7 +1265,13 @@ class Response(typing.Generic[T]):
     output: T = None
     detour: jobflow.Flow | Job | list[Job] | list[jobflow.Flow] = None
     addition: jobflow.Flow | Job | list[Job] | list[jobflow.Flow] = None
-    replace: jobflow.Flow | Job | list[Job] | list[jobflow.Flow] = None
+    replace: (
+        jobflow.Flow
+        | Job
+        | list[Job]
+        | list[jobflow.Flow]
+        | dict[Any, jobflow.Job | jobflow.Flow]
+    ) = None
     stored_data: dict[Hashable, Any] = None
     stop_children: bool = False
     stop_jobflow: bool = False
@@ -1312,8 +1318,13 @@ class Response(typing.Generic[T]):
             return isinstance(x, Job | Flow)
 
         should_replace = is_job_or_flow(job_returns)
-        if job_returns and isinstance(job_returns, (list, tuple)):
-            should_replace = all(is_job_or_flow(resp) for resp in job_returns)
+        if job_returns:
+            if isinstance(job_returns, (list, tuple)):
+                should_replace = all(is_job_or_flow(resp) for resp in job_returns)
+            elif isinstance(job_returns, dict):
+                should_replace = all(
+                    is_job_or_flow(resp) for resp in job_returns.values()
+                )
 
         if should_replace:
             job_returns = Response(replace=job_returns)
@@ -1330,6 +1341,13 @@ class Response(typing.Generic[T]):
         if isinstance(job_returns, (list, tuple)):
             # check that a Response object is not given as one of many outputs
             for resp in job_returns:
+                if isinstance(resp, Response):
+                    raise ValueError(
+                        "Response cannot be returned in combination with other outputs."
+                    )
+        elif isinstance(job_returns, dict):
+            # check that a Response object is not given as one of many outputs
+            for resp in job_returns.values():
                 if isinstance(resp, Response):
                     raise ValueError(
                         "Response cannot be returned in combination with other outputs."
@@ -1393,7 +1411,7 @@ def store_inputs(inputs: Any) -> Any:
 
 
 def prepare_replace(
-    replace: jobflow.Flow | Job | list[Job],
+    replace: jobflow.Flow | Job | list[Job] | dict[Any, jobflow.Job | jobflow.Flow],
     current_job: Job,
 ) -> jobflow.Flow:
     """
@@ -1418,6 +1436,10 @@ def prepare_replace(
         The updated flow.
     """
     from jobflow.core.flow import Flow
+
+    if isinstance(replace, dict):
+        output = {k: j.output for k, j in replace.items()}
+        replace = Flow(jobs=list(replace.values()), output=output)
 
     if isinstance(replace, (list, tuple)):
         replace = Flow(jobs=replace)
--- a/src/jobflow/core/job.py
+++ b/src/jobflow/core/job.py
@@ -1268,8 +1268,8 @@ class Response(typing.Generic[T]):
     replace: (
         jobflow.Flow
         | Job
-        | list[Job]
-        | list[jobflow.Flow]
+        | tuple[jobflow.Job | jobflow.Flow]
+        | list[jobflow.Job | jobflow.Flow]
         | dict[Any, jobflow.Job | jobflow.Flow]
     ) = None
     stored_data: dict[Hashable, Any] = None
@@ -1318,10 +1318,10 @@ class Response(typing.Generic[T]):
             return isinstance(x, Job | Flow)
 
         should_replace = is_job_or_flow(job_returns)
-        if job_returns:
-            if isinstance(job_returns, (list, tuple)):
+        if job_returns is not None:
+            if isinstance(job_returns, (list, tuple)) and len(job_returns) > 0:
                 should_replace = all(is_job_or_flow(resp) for resp in job_returns)
-            elif isinstance(job_returns, dict):
+            elif isinstance(job_returns, dict) and len(job_returns) > 0:
                 should_replace = all(
                     is_job_or_flow(resp) for resp in job_returns.values()
                 )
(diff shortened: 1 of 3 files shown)


vineetbansal:
I just saw that @gpetretto submitted a [small fix](https://github.com/materialsproject/jobflow/pull/848) to the [previous PR](https://github.com/materialsproject/jobflow/pull/833). That particular issue does not exist in this PR since everything is inside the `if` block, but let me add a small check in the unit test for this as well...

src/jobflow/core/job.py
```diff
@@ -1312,8 +1318,13 @@ def is_job_or_flow(x):
             return isinstance(x, Job | Flow)
 
         should_replace = is_job_or_flow(job_returns)
-        if job_returns and isinstance(job_returns, (list, tuple)):
-            should_replace = all(is_job_or_flow(resp) for resp in job_returns)
+        if job_returns:
```
gpetretto: Hi @vineetbansal, there was this issue #855 just reported that will also affect the changes in this PR. As @davidwaroquiers was suggesting,here maybe it can be `if isinstance(job_returns, (list, tuple, dict)) and len(job_returns) > 0:` instead of just `if job_returns:`, so it is fixed for all these cases.
Or if you have any better check that covers cases where `if job_returns` will fail.

src/jobflow/core/job.py
```diff
@@ -1334,6 +1345,13 @@ def is_job_or_flow(x):
                     raise ValueError(
                         "Response cannot be returned in combination with other outputs."
                     )
+        elif isinstance(job_returns, dict):
```
gpetretto: I am not sure I see the point of this check. why a dict or `Response` is a particular use case that deserves a check? What if the job returns a list/tuple of Response? a numpy array of Response? an arbitrary object that contains a Response?
Is there a reason why the dictionary is a special case?

src/jobflow/core/job.py
```diff
@@ -1419,6 +1437,10 @@ def prepare_replace(
     """
     from jobflow.core.flow import Flow
 
+    if isinstance(replace, dict):
+        output = {k: j.output for k, j in replace.items()}
+        replace = Flow(jobs=list(replace.values()), output=output)
```
gpetretto, line 1446: This is not strictly a comment about this PR, but rather the previous one. I think it makes sense that here the output of the Flow is set to `{k: j.output for k, j in replace.items()}`. So, wouldn't it make sense that also for the case below, when `replace` is a list/tuple, the output of the created flow is set to `[j.output for j in replace]`?

tests/core/test_flow.py
```diff
@@ -1390,10 +1390,44 @@ def add_distributed_list(list_a):
...
+
+    all_outputs = [
+        response.output
+        for index_to_response in results.values()
+        for response in index_to_response.values()
+    ]
+    assert 5 in all_outputs  # sum
+    assert 1 in all_outputs  # diff
+
+
```
gpetretto, line 1441: maybe add a test for a job returning an empty dictionary?

src/jobflow/core/job.py
```diff
@@ -1265,7 +1265,13 @@ class Response(typing.Generic[T]):
     output: T = None
     detour: jobflow.Flow | Job | list[Job] | list[jobflow.Flow] = None
     addition: jobflow.Flow | Job | list[Job] | list[jobflow.Flow] = None
-    replace: jobflow.Flow | Job | list[Job] | list[jobflow.Flow] = None
+    replace: (
+        jobflow.Flow
+        | Job
+        | list[Job]
+        | list[jobflow.Flow]
```
gpetretto: Should this also be a `list[Job | Flow]` instead of `list[Job] | list[Flow]`?

gpetretto (commented):
Hi @vineetbansal,

thanks for the addition. I have left a few small comments. In particular it would be good to address here the problem reported in #855.

src/jobflow/core/job.py
vineetbansal: Yes - makes sense.  I've added a test `test_flow.py::test_flow_no_truthiness` for this.

src/jobflow/core/job.py
vineetbansal: Yes! Added now. The slightly verbose way of handling it is to keep `mypy` happy:
```
        if isinstance(replace, tuple):
            seq_output: tuple | list = tuple(j.output for j in replace)
        else:
            seq_output = [j.output for j in replace]
```

tests/core/test_flow.py
vineetbansal: Added a check towards the end of `test_flow.py::test_job_autoreplace_dict` for an empty dict, along the same lines as `test_flow.py::test_job_autoreplace`.

src/jobflow/core/job.py
vineetbansal: Fixed

src/jobflow/core/job.py
vineetbansal: Note the comment below related to this change.

vineetbansal:
Thanks @gpetretto. I think I've addressed the concerns here. 

Some tests had to be tweaked slightly because with this change:

From:
```
    if isinstance(replace, (list, tuple)):
        replace = Flow(jobs=replace)
```
to:
```
    if isinstance(replace, (list, tuple)):
        if isinstance(replace, tuple):
            seq_output: tuple | list = tuple(j.output for j in replace)
        else:
            seq_output = [j.output for j in replace]
        replace = Flow(jobs=list(replace), output=seq_output)
```
The last `response.replace` becomes a `store_inputs` job.

However, in doing this, I noticed something that may or may not indicate something problematic. This snippet in one of the tests:

```
    test_job = Job(replace_flow, config=nopass_config)
    response = test_job.run(memory_jobstore)
    for j in response.replace:
        assert j.config.manager_config == {}
```
becomes:
```
    test_job = Job(replace_flow, config=nopass_config)
    response = test_job.run(memory_jobstore)
    for j in response.replace[:-1]:
        assert j.config.manager_config == {}
    assert response.replace[-1].config.manager_config == manager_config2
```

The fact that the last `response.replace.config.manager_config` is `manager_config2` now doesn't make sense to me. It seems to me like in returning the output of each individual job in the iterable, we may have stumbled upon an unrelated bug?

gpetretto:
Thanks @vineetbansal for the updates. I think they are fine, but apparently I cannot mark the comments as resolved.

Concerning the change in the config manager, you are absolutely right, I think it is a bug. The problem is that the `store_inputs` job (as would be any other Job obtained from the `@job` decorator) share the same instace of `JobConfig` if it is passed to the decorator itself. So, when during the tests the `job.config.manager_config` is modified multiple times, it changes for all the others jobs as well. I think this functionality is not used very often and if jobs are executed in isolated environments (like in fireworks and jobflow-remote) one would never have this issue. Still, it is a potential source of errors and it would be better to fix this.

I think that the correct solution would be to make this change:
```diff
--- a/src/jobflow/core/job.py
+++ b/src/jobflow/core/job.py
@@ -184,6 +184,7 @@ def job(
     --------
     Job, .Flow, .Response
     """
+    from copy import deepcopy

     def decorator(func):
         from functools import wraps
@@ -214,7 +215,7 @@ def job(
                         args = args[1:]

             return Job(
-                function=f, function_args=args, function_kwargs=kwargs, **job_kwargs
+                function=f, function_args=args, function_kwargs=kwargs, **deepcopy(job_kwargs)
             )
```

I briefly tested and with this the tests can go back to their original form. It makes sense to make sure that the same instances of the jobs_kwargs elements are not shared among the different instances of the jobs created from a single decorated function. @utf do you see any problem with this? or do you have any better solution?

vineetbansal:
Thanks @gpetretto - that makes sense. I've removed the last `assert` (which will change if/when this `deepcopy` fix is introduced), since it is unrelated to the change we're trying to make here. Without that `assert`, it is clear that what we're testing is that the `config` for all the replaced jobs (except the last `store_inputs` job) is empty.

I'll open up another issue on the deepcopy stuff in case it has deeper implications that we may want to investigate.

vineetbansal:
@utf - perhaps you can take a look? We're not introducing any deep changes here other than interpreting dict return of `jobs` as `replace` (list of jobs being returned are already handled as a `replace` and were approved already).

Andrew-S-Rosen:
Hi @utf, I think this one is ready to go. Many of the changes are with respect to the test suite.

Andrew-S-Rosen:
Hi @utf, next week is the last week @vineetbansal will be working with my group, so it would be great to get this reviewed and merged soon if possible. Thank you!!

utf:
Hi, thanks for this. It looks good to me.

PR #860, opened 2026-03-13
A possible issue with not deepcopying job kwargs

vineetbansal:
As we discovered when working towards another [PR](https://github.com/materialsproject/jobflow/pull/856), mutating one job's config mutates all others created by that decorator. This is because the `@job` decorator keeps a reference to `job_kwargs` instead of deep-copying it. This may or may not be an issue, and may just need explicit documentation.

This PR introduces an `xfail` test to demonstrate this behavior. The solution suggested by @gpetretto which should solve this is:

```
--- a/src/jobflow/core/job.py
+++ b/src/jobflow/core/job.py
@@ -184,6 +184,7 @@ def job(
     --------
     Job, .Flow, .Response
     """
+    from copy import deepcopy

     def decorator(func):
         from functools import wraps
@@ -214,7 +215,7 @@ def job(
                         args = args[1:]

             return Job(
-                function=f, function_args=args, function_kwargs=kwargs, **job_kwargs
+                function=f, function_args=args, function_kwargs=kwargs, **deepcopy(job_kwargs)
             )
```

### DIFF
--- a/tests/core/test_job.py
+++ b/tests/core/test_job.py
@@ -1418,7 +1418,8 @@ def test_job_magic_methods():
 
 
 @pytest.mark.xfail(
-    reason="Mutating one job's config mutates all others created by that decorator."
+    reason="Mutating one job's config mutates all others created by that decorator.",
+    strict=True,
 )
 def test_job_decorator_config_shared():
     from jobflow import JobConfig, job
--- a/tests/core/test_job.py
+++ b/tests/core/test_job.py
@@ -1415,3 +1415,30 @@ def test_job_magic_methods():
 
     # test __hash__
     assert hash(job1) != hash(job2) != hash(job3)
+
+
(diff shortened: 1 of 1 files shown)


gpetretto:
Sorry @vineetbansal, can you confirm that this was just the test to verify the bug and that the proposed solution has not been included in any other of your PRs? If that's the case I will open a PR with the proposed fix.

vineetbansal:
Hi @gpetretto - yes this is just to demonstrate the issue. The fix is not in this this PR or any other. You can add the fix to this PR and merge (removing the `xfail`) or let me know and I'll do so. Or we can close this PR if you're going to be opening up a new one.

gpetretto:
If you have the time to add this, it is fine with me. But if you are short on time let me know and I can also do that.

PR #863, opened 2026-03-27
implement local version of lenient_issubclass

gpetretto:
Following a report from @FabiPi3, the current version of pydantic (2.12.5) raises a warning that the pydantic.v1 module is not compatible with python 3.14 or later. While this has apparently been fixed in pydantic and the warning will be gone in the next release, support for the pydantic.v1 will likely be dropped in v3: https://github.com/pydantic/pydantic/issues/10033. 
I think it is worth getting rid of that import right away, considering that the usage of that module is minimal in jobflow.

The implementation is slightly different from the original (https://github.com/pydantic/pydantic/blob/8ba0d00782cc453ab9542f8932ef2b2a47614594/pydantic/v1/utils.py#L181) since I think that we don't want the `lenient_issuclass` to ever raise an error.

### DIFF
--- a/src/jobflow/core/reference.py
+++ b/src/jobflow/core/reference.py
@@ -8,9 +8,9 @@ from typing import Any
 
 from monty.json import MontyDecoder, MontyEncoder, MSONable, jsanitize
 from pydantic import BaseModel
-from pydantic.v1.utils import lenient_issubclass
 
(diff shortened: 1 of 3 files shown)


FabiPi3:
Thank you very much

utf:
Great, thanks @gpetretto

PR #885, opened 2026-08-10
Resolve Job inputs automatically in decorated flows

### LINKED ISSUE #884: Feature request: automatic .output resolving in flow-decorated functions
The following does not work because `a * b` cannot be done, as `sum_result` is a `Job` and not the resolved output. It would be nice to be able to construct flows without having to explicitly call `.output`. We can already return `Job` objects and get the `.output` automatically resolved, which is why the second example works, but the first example would be ideal to get working too.

This fails:
```python
from jobflow import job, flow, run_locally

@job
def add(a, b):
    return a + b

@job
def mult(a, b):
    return a * b

@flow
def workflow(a, b):
    sum_result = add(a, b)
    mult_result = mult(a, sum_result)
    return mult_result

output = run_locally(workflow(1, 2))
```

This works:
```python
@job
def add(a, b):
    return a + b

@job
def mult(a, b):
    return a * b

@flow
def workflow(a, b):
    sum_result = add(a, b)
    mult_result = mult(a, b)
    return mult_result, sum_result

output = run_locally(workflow(1, 2))
```

Andrew-S-Rosen:
Closes #884.

## What changed

- Convert `Job` and `Flow` values passed to jobs inside `@flow`-decorated functions into their output references automatically.
- Apply the conversion recursively to list, tuple, and dictionary inputs.
- Keep conversion limited to decorated-flow construction; runtime job execution is unchanged.
- Add regression tests for the minimal example from #884.

## Why

The decorated-flow build context already collects delayed jobs, but downstream job construction retained whole `Job` objects as inputs. This produced an invalid job argument and required users to write `.output` explicitly. Replacing these values while the decorated flow is being built makes the shorthand behave as expected without changing job construction outside `@flow`.

### DIFF
--- a/src/jobflow/core/flow.py
+++ b/src/jobflow/core/flow.py
@@ -284,11 +284,8 @@ class Flow(MSONable):
             The output of the flow. These should come from the output of one
             or more of the jobs.
         """
-        output_is_normalized = getattr(self, "_output_is_normalized", False)
-        self._output_is_normalized = False
-
         if output is not None:
-            if not output_is_normalized and contains_flow_or_job(output):
+            if contains_flow_or_job(output):
                 warnings.warn(
                     f"Flow '{self.name}' contains a Flow or Job as an output. "
                     f"Usually the Flow output should be the output of a Job or "
@@ -992,9 +989,6 @@ class DecoratedFlow(Flow):
                 stacklevel=2,
             )
         output = _normalize_job_or_flow(output)
-        # Avoid checking the entire output again in Flow.output; normalization has
-        # already replaced every nested Job and Flow.
-        self._output_is_normalized = True
 
         super().__init__(name=name, jobs=children_list, output=output)
 
--- a/src/jobflow/core/flow.py
+++ b/src/jobflow/core/flow.py
@@ -932,16 +932,21 @@ def get_flow(
 
 
 def _normalize_job_or_flow(value):
-    """Replace Jobs and Flows in nested containers with their outputs."""
+    """Replace Jobs and Flows at the top level of a value with their outputs."""
+    def normalize_item(item):
+        if isinstance(item, (jobflow.Job, jobflow.Flow)):
+            return item.output
+        return item
+
     if isinstance(value, (jobflow.Job, jobflow.Flow)):
         return value.output
     if isinstance(value, list):
-        return [_normalize_job_or_flow(item) for item in value]
+        return [normalize_item(item) for item in value]
     if isinstance(value, tuple):
-        return tuple(_normalize_job_or_flow(item) for item in value)
+        return tuple(normalize_item(item) for item in value)
     if isinstance(value, dict):
         return {
-            _normalize_job_or_flow(key): _normalize_job_or_flow(item)
+            normalize_item(key): normalize_item(item)
             for key, item in value.items()
         }
     return value
--- a/src/jobflow/core/flow.py
+++ b/src/jobflow/core/flow.py
@@ -284,8 +284,11 @@ class Flow(MSONable):
             The output of the flow. These should come from the output of one
             or more of the jobs.
         """
+        output_is_normalized = getattr(self, "_output_is_normalized", False)
+        self._output_is_normalized = False
+
         if output is not None:
-            if contains_flow_or_job(output):
+            if not output_is_normalized and contains_flow_or_job(output):
                 warnings.warn(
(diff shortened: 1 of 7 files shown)


Andrew-S-Rosen:
@utf --- could I get your review when you have a moment? Thanks.

src/jobflow/core/flow.py
```diff
@@ -931,6 +931,27 @@ def get_flow(
     return flow
 
 
+def _normalize_job_or_flow(value):
```
gpetretto: I think that "normalize" in the name is not really conveying the meaning of what the function is doing. It may be somewhat clear if one has in mind the `@flow` decorator, but it seems that the name may trick someone into thinking that this has a more general purpose. Maybe the name can be more explicit in what the method does?

Also, since this is used both in the flow and job modules, maybe it can be in the utils?

src/jobflow/core/job.py
```diff
@@ -213,6 +213,12 @@ def get_job(*args, **kwargs) -> Job:
                         f = met
                         args = args[1:]
 
+            if _current_flow_context.get() is not None:
+                args = tuple(_normalize_job_or_flow(arg) for arg in args)
```
gpetretto: can't this just be 
```python
args = _normalize_job_or_flow(args)
```
?

src/jobflow/core/job.py
```diff
@@ -213,6 +213,12 @@ def get_job(*args, **kwargs) -> Job:
...
+                kwargs = {
+                    key: _normalize_job_or_flow(value) for key, value in kwargs.items()
+                }
```
gpetretto: As above, can't this just be 
```python
kwargs = _normalize_job_or_flow(kwargs)
```
?

src/jobflow/core/job.py
```diff
@@ -626,7 +637,13 @@ def run(self, store: jobflow.JobStore, job_dir: Path = None) -> Response:
         if bound is not None and not isinstance(bound, types.ModuleType):
             function = types.MethodType(function, bound)
 
-        response = function(*self.function_args, **self.function_kwargs)
+        children: list[Job | Flow] = []
+        with flow_build_context(children):
```
gpetretto: I would consider this point a bug. The main point of `flow_build_context` should be to distinguish a standard Flow from a DecoreatedFlow when the Flow is being created, while here it is used at runtime, which I don't think fits what it was designed for.
The immediate consequence of entering this context manager here is that any Flow dynamically generated inside this job would be considered as a `DecoratedFlow`. As an example I have explicitly verifiedthat even Flows coming from makers have their values replaced with the `.output` attribute automatically. Whether this is desirable could be a point to be discussed, but I think it goes beyond the scope of this PR. 

Also, I suspect there may be other unwanted consequences that are difficult to predict without an extensive testing of different use cases. And it may create problems for this other #850 idea that we were trying to develop.

I quickly tested without this and it does not seem needed to run your example in #884. Why is this needed? Can this part be reverted? I think it will not be possible to determine in this point of the code if the current job was actually belonging to a standard Flow or a DecoratedFlow. For sure not in jobflow-remote.

gpetretto (commented):
Hi @Andrew-S-Rosen,
I think this would be interesting. I have left a few cosmetic comments and one with a major concern. 
In general, I don't expect this to conflict with #870

src/jobflow/core/job.py
Andrew-S-Rosen: I think this was me trying to resolve the somewhat contrived example below, which does not work otherwise:

```python
from jobflow import flow, job, run_locally


@job
def double(number):
    return number * 2


@job
def increment(number):
    return number + 1


@job
def process_numbers(numbers):
    results = []

    for number in numbers:
        doubled = double(number)
        incremented = increment(doubled.output)
        results.append(incremented)

    return results


@flow
def number_processing_workflow(numbers):
    return process_numbers(numbers)


workflow = number_processing_workflow([1, 2, 3])
run_locally(workflow, ensure_success=True)
```
where `numbers` is not necessarily known _a priori_. But I see the problems you are raising.

Andrew-S-Rosen:
@gpetretto Thank you very much for the helpful review! I have addressed your comments. The buggy code you pointed out has since been removed entirely. There was an edge case I was trying to resolve (mentioned in https://github.com/materialsproject/jobflow/pull/885#discussion_r3823825873), but it is different than #884 and probably not sensible anyway. I have streamlined this PR to focus on the main issue at hand.

PR #888, opened 2026-08-11
Normalize flow outputs and retain dynamic subflow jobs

Andrew-S-Rosen:
## Summary

- Normalize nested `Job` and `Flow` values returned by `@flow` into output references.
- Collect all jobs created while a job function executes when its return value defines a dynamic subflow.
- Preserve the full dynamic subflow as the replacement flow instead of retaining only the returned terminal job.
- Add regression tests for nested decorated-flow outputs and multi-job dynamic subflows.

## Why

Downstream workflow adapters currently need to wrap decorated flow outputs recursively and compose `@job` with `@flow` to prevent intermediate jobs in dynamic subflows from being lost. This behavior is general Jobflow composition logic and belongs upstream.

Together with #885, this allows Quantum-Accelerators/quacc#3405 to remove its Jobflow-specific normalization and subflow patches from `src/quacc/wflow_tools/decorators.py`.

## Additional dependencies introduced

None.

## Validation

- `PYTHONPATH=src python -m pytest tests/core/test_flow_decorator.py tests/core/test_job.py -q` — 39 passed, 1 xfailed
- `python -m ruff check --ignore PLR0917 src/jobflow/core/flow.py src/jobflow/core/job.py tests/core/test_flow_decorator.py tests/core/test_job.py` — passed
- `git diff --check` — passed

## Checklist

- [x] Tests have been added for the new behavior.
- [x] Focused linting and tests pass.

### DIFF
--- a/src/jobflow/core/flow.py
+++ b/src/jobflow/core/flow.py
@@ -931,6 +931,22 @@ def get_flow(
     return flow
 
 
+def _normalize_flow_output(value):
+    """Replace Jobs and Flows in nested containers with their outputs."""
+    if isinstance(value, (jobflow.Job, jobflow.Flow)):
+        return value.output
+    if isinstance(value, list):
+        return [_normalize_flow_output(item) for item in value]
+    if isinstance(value, tuple):
+        return tuple(_normalize_flow_output(item) for item in value)
+    if isinstance(value, dict):
+        return {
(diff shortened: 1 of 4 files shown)


Andrew-S-Rosen:
Superseded by #885. The relevant changes and regression tests have been consolidated there using a single shared normalization helper.

